PR #42, opened 2021-08-03
Dropdown for topic selection

shinyichen:
* Dropdown for topic selection (not functional), with keyboard accessibility. 
* Some style changes on landing page

Screen Shot 2021-08-03 at 3 13 34 PM


src/components/Dropdown/DropdownList.tsx
```diff
@@ -56,21 +57,98 @@ export const DropdownList = (props: IDropdownListProps): ReactElement => {
...
+        return;
+      case 'Tab':
+        close();
+        return;
+      default:
+        return;
+    }
+  };
+
+  const focusItem = (index: number) => {
```
thostetler, line 140: This is fine as is, however I'm not a huge fan of mutating a parameter in a function.  We should use a local variable to store any updates, just as a general rule.
I changed it to a ternary, but you can modify however you'd like:
```js
const focusItem = (index: number) => {
  const numItems = items.length;
  const idx = index >= numItems ? 0 : index < 0 ? numItems - 1 : index;
  document.getElementById(`${items[idx].id}`).focus();
};
```

src/components/Dropdown/DropdownList.tsx
```diff
@@ -56,21 +57,98 @@ export const DropdownList = (props: IDropdownListProps): ReactElement => {
...
+  };
+
+  /* selected item */
+  const handleSelect = (item: ItemType) => {
+    onSelect(item.id);
+    close();    
+  };
+
+  /* keydown on item */
+  const handleItemKeyDown = (e: KeyboardEvent, item: ItemType, index: number) => {
```
thostetler, line 111: This is okay, I just did a couple edits to make it more concise:
```js
const handleItemKeyDown = (e: KeyboardEvent, item: ItemType, index: number) => {
  switch (e.key) {
    case 'Enter':
    case 'Space':
      e.preventDefault();
      return handleSelect(item);
    case 'ArrowUp':
      e.preventDefault();
      return focusItem(index - 1);
    case 'ArrowDown':
      e.preventDefault();
      return focusItem(index + 1);
    case 'Escape':
    case 'Tab':
      return close();
  }
};
```
With a switch statement we can flow down to subsequent ones if the action is the same, so thats what I did with the 'Enter'.  Also, moved the return to the line above just to clean up a little.

src/components/Dropdown/DropdownList.tsx
```diff
@@ -56,21 +57,98 @@ export const DropdownList = (props: IDropdownListProps): ReactElement => {
...
-  const handleSelect = useCallback(
-    (item: ItemType) => {
-      return () => {
-        onSelect(item.id);
-      };
-    },
-    [onSelect],
-  );
+  /* keydown on dropdown */
+  const handleKeyDown = (e: KeyboardEvent) => {
```
thostetler, line 81: I mentioned the same below, did a similar update here:

```js
const handleKeyDown = (e: KeyboardEvent) => {
  switch (e.key) {
    case 'Enter':
    case 'Space':
      e.preventDefault();
      return handleClick();
    case 'Escape':
      return close();
    case 'ArrowDown':
      e.preventDefault();
      open();
      return focusItem(0);
  }
};
```

src/components/Dropdown/DropdownList.tsx
```diff
@@ -56,21 +57,98 @@ export const DropdownList = (props: IDropdownListProps): ReactElement => {
     ],
   });
 
+  const close = () => {
```
thostetler, line 60: This is okay, however when you're essentially "watching" for changes to a state variable (in this case `visible`) its best to do it explicitly so you always stay up-to-date.  I would move the `onClosed()` and `onExpanded()` into a `useEffect` that watches for changes to `visible` so they stay in sync.
For example:

```js
// call our callbacks when visible changes
useEffect(() => (visible ? onExpanded() : onClosed()), [visible]);

// helpers for setting visibility
const close = () => visible && setVisible(false);
const open = () => !visible && setVisible(true);

/* Click on dropdown */
const handleClick = () => {
  setVisible(!visible);
};
```

src/components/NavBar/AboutDropdown.tsx
```diff
@@ -1,53 +1,50 @@
...
-    element: <ItemLink href="/about/careers"> Careers@ADS</ItemLink>,
-  },
-];
+    id: 'about-careers',
+    path: '/about/careers',
+    label: 'Careers@ADS'
+  }
+]
+
+const items: ItemType[] = (() => {
```
thostetler, line 37: This one I wasn't sure about, you're trying to map it down to only include `id` and `label`? 
I think you could get rid of the IIFE, since really this is a cheap enough operation to have directly in the function, you can just move it like this:

```js
export const AboutDropdown = (): ReactElement => {
  const handleSelect = (id: string) => {
    window.open(menu.find((item) => id === item.id).path, '_blank', 'noopener,noreferrer');
  };

  // map items to remove unnessary keys
  const items = menu.map(({ id, label }) => ({ id, label }));

  return (
    <DropdownList
      label="About"
...
```

thostetler: If you were concerned with the operation being too expensive or if this component were to render tons of times then I think an optimization would be warranted.

src/components/NavBar/TopicDropdown.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <span className="inline-block align-baseline p-1.5">{label}</span>
+      <ChevronDownIcon className="inline m-1.5 w-4 h-4" />
+    </div>
+  );
+
+  return (
+    <DropdownList
+      label={getLabelNode(selectedTopic.label)}
+      items={topics}
+      onSelect={(topicId: string) => {
```
thostetler, line 49: One small nit, instead of these functions inside the props, try to use handlers (i.e. `handleOnSelect`).  When you have a lambda type function like this it gets re-evaluated at every render.

src/components/Dropdown/DropdownBasic.tsx
```diff
@@ -52,7 +53,49 @@ export const DropdownBasic = (props: IDropdownBasicProps): ReactElement => {
...
+    visible || setVisible(true);
+  }
+
+  const close = () => {
+    !visible || setVisible(false);
+  }
+
+  /* keydown on dropdown */
+  const handleKeyDown = (e: KeyboardEvent) => {
+    switch (e.key) {
```
thostetler, line 66: As I mention below, we can make this switch statement a little more concise

src/components/Dropdown/DropdownBasic.tsx
```diff
@@ -52,7 +53,49 @@ export const DropdownBasic = (props: IDropdownBasicProps): ReactElement => {
     setVisible(false);
   };
 
-  const containerClasses = clsx('z-50 bg-white', classes.container, {
+  const open = () => {
+    visible || setVisible(true);
```
thostetler, line 57: I think you want to do an `&&` here, since the `||` is going to only ever run `setVisible()` when `visible === false`

We can change this code, as below:

```js
// helpers for setting visibility
const close = () => visible && setVisible(false);
const open = () => !visible && setVisible(true);
```

src/styles/styles.css
```diff
@@ -0,0 +1,34 @@
...
+  }
+  h4 {
+    @apply text-xl;
+  }
+
+  .default-text-color {
+    @apply text-gray-600;
+  }
+
+  .defatul-bg-color {
```
thostetler, line 23:
```suggestion
  .default-bg-color {
```

thostetler (changes requested):
This looks good! I like the dropdown, I did notice there is a bit of a gap, at least for me:
image
I'm sure you noticed that too though.  

I had some areas I thought could use some updating and a couple tips/suggestions.  Please let me know if you disagree with any proposed changes.

src/styles/styles.css
```diff
@@ -0,0 +1,34 @@
...
+    @apply bg-white;
+  }
+
+  .navbar-text-color {
```
thostetler, line 27: instead of adding `navbar` colors here, you can also add this css module inside the `navbar` directory and include it there directly.  (details here: https://nextjs.org/docs/basic-features/built-in-css-support#adding-component-level-css)
Just uses css-modules, so you don't have to add them globally here.

src/components/NavBar/AboutDropdown.tsx
shinyichen: I removed `element` for the following reasons:
1) Trying to maintain a consistent look and feel with the dropdown, so only let parent handle the click.
2) Accessibility in the dropdown using keyboard was very tricky

shinyichen: The mapping was meant to map them to `ItemType` type.

src/components/Dropdown/DropdownList.tsx
thostetler: Yeah we could keep it, but since you aren't really returning a value other than `undefined` I think we can leave it off and return after the switch.

src/components/NavBar/AboutDropdown.tsx
thostetler: Okay that makes sense, you could also update `ItemType` to remove `element` and add `path` or whatever.
I think its only being used in the navbar right now right?

src/components/Dropdown/DropdownBasic.tsx
thostetler: Technically that's true, on second look at this! :laughing:  
I would say, looking at this again we should scrap doing this type of short-circuiting since React will optimize a state update anyway.
We can just do `open = () => setVisible(true);` and if its already `true` react won't re-render.

shinyichen:
> This looks good! I like the dropdown, I did notice there is a bit of a gap, at least for me:
> image
> I'm sure you noticed that too though.
> 
> I had some areas I thought could use some updating and a couple tips/suggestions. Please let me know if you disagree with any proposed changes.

src/components/Dropdown/DropdownList.tsx
shinyichen: Updated

src/components/Dropdown/DropdownList.tsx
shinyichen: Updated

src/components/Dropdown/DropdownList.tsx
shinyichen: Updated

src/components/Dropdown/DropdownList.tsx
shinyichen: Updated

src/components/NavBar/AboutDropdown.tsx
shinyichen: Updated

src/components/NavBar/TopicDropdown.tsx
shinyichen: Done

src/components/Dropdown/DropdownBasic.tsx
shinyichen: Done

src/components/Dropdown/DropdownBasic.tsx
shinyichen: Updated

src/styles/styles.css
shinyichen: Updated

src/styles/styles.css
shinyichen: Done

src/components/NavBar/TopicDropdown.tsx
```diff
@@ -0,0 +1,65 @@
...
+  },
+];
+
+export const TopicDropdown = (): ReactElement => {
+  const [selectedTopic, setSelectedTopic] = useState<ItemType>(topics[0]);
+
+  const getLabelNode = (label: string) => (
+    <div
+      id="topicSelector"
+      className={`${styles['navbar-bg-color']} ${styles['navbar-text-color']} flex items-center justify-between w-64 border border-gray-50 border-opacity-50 rounded-sm cursor-pointer`}
```
thostetler, line 39: We may want to use `clsx` just to clean up these classes instead of using template strings.  But not a show stopper 😃

thostetler (approved):
LGTM, I made one small comment on swapping out the navbar styles for clsx.  But we can change that later, let's merge this for now

PR #43, opened 2021-08-09
Make changing the topic/theme functional

shinyichen:


src/store/types.ts
```diff
@@ -1,11 +1,14 @@
...
+  theme: Theme;
 }
 
 export enum AppEvent {
   SET_USER = 'SET_USER',
+  SET_THEME = 'SET_THEME',
 }
 
-export type Action = { type: AppEvent.SET_USER; payload: IAppState['user'] };
+export type Action = { type: AppEvent; payload: IAppState['user'] | IAppState['theme'] };
```
thostetler, line 14: Lets switch to using discriminated unions here, it'll type better (https://basarat.gitbook.io/typescript/type-system/discriminated-unions#discriminated-union)


```suggestion
export type Action = 
| { type: AppEvent.SET_USER; payload: IAppState['user'] }
| { type: AppEvent.SET_THEME; payload: IAppState['theme'] }
```

thostetler: Once you make this change you should be able to go back to the areas you've used them and remove any casting (i.e. `as Theme`)

src/pages/paper-form.tsx
```diff
@@ -1,7 +1,18 @@
...
+import { Theme } from '@types';
 import { NextPage } from 'next';
+import { useRouter } from 'next/router';
 import React from 'react';
+import { useEffect } from 'react';
 
 const PaperForm: NextPage = () => {
+  const Router = useRouter();
+  const { state } = useAppCtx()
+  useEffect(() => {
```
thostetler, line 11: is this just to redirect us back to the modern form if the theme is something other than astrophysics?

I wonder if we should maybe move this logic upward into `_app` and possible into a hook that does this check, hmm

shinyichen: Right. If current path is classic or paper form, but theme is not astrophysics. I'll look into this, let me know if you already have some ideas in mind.

shinyichen: Moved the redirect logic to `_app`

thostetler:
Nice work on this, it looks good to me -- merge when you're ready.

PR #54, opened 2021-09-03
Disable next telemetry, separate dev from dist (prod), improved dockerfile

marblestation:
- All built file are now saved into 'dist' instead of '.next' and 'dist'
- Running the dev server does not overwrite the production build in the 'dist' directory, but it uses a 'dev' directory
- Disable next telemetry when building or starting the dev server
- Updated README with more install/running details
- Added unmet peer dependencies
- Added the possibility to run cypress from the docker container


thostetler (approved):
This LGTM, I think we can merge

PR #57, opened 2021-09-09
Add paper form

thostetler:
Also makes several other small changes to bibstemPicker, api, and how tokens are handled


thostetler:
Going to close for the moment and break apart

thostetler:
@shinyichen scope grew a bit in this PR, please let me know if you want to discuss the changes during your review :)

src/components/Button/Button.tsx
```diff
@@ -1,14 +1,46 @@
...
+enum ButtonSize {
+  XS = 'xs',
+  SM = 'sm',
+  MD = 'md',
+  LG = 'lg',
+  XL = 'xl',
+}
+
+export interface IButtonProps extends DetailedHTMLProps<ButtonHTMLAttributes<HTMLButtonElement>, HTMLButtonElement> {
+  variant?: `${ButtonVariant}`;
```
shinyichen, line 18: Is this the same thing as `variant?: ButtonVariant;`?

thostetler: This is a template literal type, it'll enumerate the `enum` into a union of strings.  

https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html

Pretty useful, and it'll work with any iterable type I think

thostetler: To actually answer your question though, it's not the exactly the same, if you do just `ButtonVariant` it won't suggest a string, you would have to do `variant={ButtonVariant.Primary}`

Whereas this will generate a type that looks like: `primary | secondary | foo`
So you can then just do:

`variant="primary"` and it'll suggest it to you

src/pages/api/paperform/[id].ts
```diff
@@ -0,0 +1,29 @@
...
+  } = req;
+
+  const adsapi = new Adsapi({ token: req.session.userData.access_token });
+
+  try {
+    const controller = new PaperFormController(id, body, adsapi);
+    const query = await controller.getQuery();
+    res.writeHead(302, { Location: `/search?${query}` });
+  } catch (e) {
+    res.writeHead(500, { Location: '/error/server' });
```
Comment (shinyichen) [inline src/pages/api/paperform/[id].ts:25]: Where does `/error/server` go to? Is it something we can see inside `/pages` ?

Comment (thostetler) [reply src/pages/api/paperform/[id].ts:25]: Yeah for now, I didn't make a page.  Gonna have to go back over things to try to think about error handling here

shinyichen (approved):

PR #60, opened 2021-09-16
Results List Updates

shinyichen:
* Removed show/hide abstract button, use individual item's show/hide abstract button
* Add some logic to list actions (show all, show none, exclude, limited to, exclude)


thostetler (approved):
LGTM

PR #63, opened 2021-09-24
Graphics page

shinyichen:
### Check whether to enable or disable the graphic link requires a query. But with current implementation, each time the menu is re-rendered (queried again) when the page changes. Is it possible to change the implementation so that we only update the content and not the menu?

Screen Shot 2021-09-24 at 5 15 39 PM

### DIFF
--- a/src/pages/abs/[id]/graphics.tsx
+++ b/src/pages/abs/[id]/graphics.tsx
@@ -7,7 +7,7 @@ import { AbstractSideNav } from '@components';
 import { normalizeURLParams } from '@utils';
 import { IADSApiGraphicsResponse } from '@api/lib/graphics/types';
 import Link from 'next/link';
-
+import Image from 'next/image';
 interface IGraphicsPageProps {
   graphics: IADSApiGraphicsResponse;
   originalDoc: IDocsEntity;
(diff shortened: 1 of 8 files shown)


src/pages/abs/[id]/graphics.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const params: IADSApiGraphicsParams = {
+    bibcode: query.id,
+  };
+  const adsapi = new AdsApi({ token: userData.access_token });
+  const result = await adsapi.graphics.query(params);
+  const originalDoc = await getOriginalDoc(adsapi, query.id);
+
+  if (result.isErr()) {
+    return { props: { graphics: [], originalDoc, error: result.error } };
+  } else if (result.value.Error) {
```
Comment (thostetler) [inline src/pages/abs/[id]/graphics.tsx:85]: we should move this logic back into the service, so an `Error` gets returned as an error result.

thostetler:
So far I think this looks pretty good! 
The only thing, as I mentioned, I think we should move that error check into the api handler

src/pages/abs/[id]/graphics.tsx
```diff
@@ -0,0 +1,95 @@
...
+            <div className="flex flex-wrap">
+              {graphics.figures.map((figure, index) => {
+                return (
+                  <div
+                    key={index}
+                    className="flex flex-col items-center justify-between m-2 p-2 border-2 border-gray-100 rounded-lg"
+                  >
+                    <Link href={figure.images[0].highres}>
+                      <a target="_blank" rel="noreferrer noopener">
+                        <img src={figure.images[0].thumbnail} className="p-5" alt={figure.figure_label}></img>
```
Comment (thostetler) [inline src/pages/abs/[id]/graphics.tsx:45]: Any reason not to use Next's `<Image/>` tag here?

src/pages/abs/[id]/graphics.tsx
Comment (shinyichen) [reply src/pages/abs/[id]/graphics.tsx]: Resolved

src/pages/abs/[id]/graphics.tsx
Comment (shinyichen) [reply src/pages/abs/[id]/graphics.tsx]: Changed to `<Image>`

PR #64, opened 2021-09-29
Abstract page - show more|less authors, affiliations, use layout file

shinyichen:
* option to show more|less authors
* option to show|hide  affiliations
* use layout files for the abs pages
Screen Shot 2021-09-29 at 3 15 13 PM

Screen Shot 2021-09-29 at 3 16 15 PM

### DIFF
--- a/src/pages/abs/[id]/abstract.tsx
+++ b/src/pages/abs/[id]/abstract.tsx
@@ -9,20 +9,27 @@ import { normalizeURLParams } from 'src/utils';
 import Link from 'next/link';
 import Image from 'next/image';
 import { createUrlByType } from '@components/AbstractSources/linkGenerator';
+import clsx from 'clsx';
 
 export interface IAbstractPageProps {
   doc?: IDocsEntity;
   error?: Error;
+  userData: IUserData;
+  params: {
+    q: string;
+    fl: string[];
+    sort: SolrSort[];
+  };
 }
 
 const MAX_AUTHORS = 20;
 
 const AbstractPage: NextPage<IAbstractPageProps> = (props: IAbstractPageProps) => {
-  const { doc, error } = props;
+  const { doc, error, userData, params } = props;
 
   const [showNumAuthors, setShowNumAuthors] = useState<number>(MAX_AUTHORS);
 
-  const [showAff, setShowAff] = useState<boolean>(false);
+  const [aff, setAff] = useState({ show: false, data: [] as string[] });
 
   // onComponentDidMount
   useEffect(() => {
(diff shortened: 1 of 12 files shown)


src/pages/abs/[id]/abstract.tsx
```diff
@@ -4,33 +4,102 @@ import { abstractPageNavDefaultQueryFields } from '@components/AbstractSideNav/m
...
+  const [showNumAuthors, setShowNumAuthors] = useState<number>(MAX_AUTHORS);
+
+  const [aff, setAff] = useState({ show: false, data: [] as string[] });
+
+  // onComponentDidMount
+  useEffect(() => {
+    if (showNumAuthors > doc.author.length) {
+      setShowNumAuthors(doc.author.length);
+    }
+  }, []);
```
Comment (thostetler) [inline src/pages/abs/[id]/abstract.tsx:40]: Doc should probably be a dep here, even though we'll be rerendering anyway

src/pages/abs/[id]/abstract.tsx
```diff
@@ -4,33 +4,102 @@ import { abstractPageNavDefaultQueryFields } from '@components/AbstractSideNav/m
...
+  const handleShowAllAuthors = () => {
+    setShowNumAuthors(doc.author.length);
+  };
+
+  const handleShowLessAuthors = () => {
+    setShowNumAuthors(Math.min(doc.author.length, MAX_AUTHORS));
+  };
+
+  const handleShowAff = () => {
+    const adsapi = new AdsApi({ token: userData.access_token });
```
Comment (thostetler) [inline src/pages/abs/[id]/abstract.tsx:51]: You should switch to use the hook probably
```suggestion
    const { api } = useAPI();
```

src/pages/abs/[id]/abstract.tsx
```diff
@@ -4,33 +4,102 @@ import { abstractPageNavDefaultQueryFields } from '@components/AbstractSideNav/m
...
+    if (aff.data.length === 0) {
+      params.fl = ['aff'];
+      void adsapi.search.query(params).then((res) => {
+        if (res.isOk()) {
```
Comment (thostetler) [inline src/pages/abs/[id]/abstract.tsx:55]: Make sure you take advantage of the `Result` it makes you handle any errors, you can use `.match()` to handle each case

thostetler (changes requested):
Looks pretty good to me, a couple changes and then few free to merge

src/pages/abs/[id]/abstract.tsx
Comment (shinyichen) [reply src/pages/abs/[id]/abstract.tsx]: Added doc as dependency

src/pages/abs/[id]/abstract.tsx
Comment (shinyichen) [reply src/pages/abs/[id]/abstract.tsx]: Updated using suggested

src/pages/abs/[id]/abstract.tsx
Comment (shinyichen) [reply src/pages/abs/[id]/abstract.tsx]: Using `.match()`

PR #65, opened 2021-09-30
Error handling for invalid document

shinyichen:
Redirect to 404 page when document is not found

### DIFF
--- a/src/pages/abs/[id]/similar.tsx
+++ b/src/pages/abs/[id]/similar.tsx
@@ -1,10 +1,8 @@
 import AdsApi, { IADSApiSearchParams, IDocsEntity, IUserData } from '@api';
-import { abstractPageNavDefaultQueryFields } from '@components/AbstractSideNav/model';
 import { AbsLayout } from '@components/Layout/AbsLayout';
 import { GetServerSideProps, NextPage } from 'next';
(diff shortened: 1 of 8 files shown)


src/pages/abs/[id]/similar.tsx
```diff
@@ -55,14 +57,16 @@ export const getServerSideProps: GetServerSideProps<ICitationsPageProps> = async
...
-  }
-
-  return {
-    props: {
-      docs: result.value.docs,
-      originalDoc,
-    },
-  };
+  return !originalDoc
+    ? { props: { error: 'Document not found' } }
```
Comment (thostetler) [inline src/pages/abs/[id]/similar.tsx:61]: Looks like we can shortcut this a bit by doing this: https://nextjs.org/blog/next-10#notfound-support

`{ notFound: true }` <-- will automatically show the 404 error page

Comment (thostetler) [reply src/pages/abs/[id]/similar.tsx:61]: We'll want to check on what kind of error we're getting and act accordingly

thostetler (commented):
LGTM -- Take a look at https://nextjs.org/blog/next-10#notfound-support, and maybe we can make this error handling a little cleaner.

shinyichen:
@thostetler Now separating the different errors, please see if this enough. Thanks.

PR #72, opened 2021-10-05
Metrics page

shinyichen:
* Metrics page
* Fixed some build errors
* Made abstract nav data render on server side

### DIFF
--- a/src/api/lib/metrics/types.ts
+++ b/src/api/lib/metrics/types.ts
@@ -1,0 +1,104 @@
+export interface IADSApiMetricsParams {
+  bibcode: string;
+}
+
+export interface ICitationsHistogram {
+  'nonrefereed to nonrefereed': {
+    year: number;
+  };
+  'nonrefereed to nonrefereed normalized': {
+    year: number;
+  };
+  'nonrefereed to refereed': {
+    year: number;
+  };
+  'nonrefereed to refereed normalized': {
+    year: number;
+  };
+  'refereed to nonrefereed': {
+    year: number;
+  };
+  'refereed to nonrefereed normalized': {
+    year: number;
+  };
+  'refereed to refereed': {
+    year: number;
+  };
+  'refereed to refereed normalized': {
+    year: number;
+  };
+}
+
+export interface IReadsHistogram {
+  'all reads': {
+    year: number;
+  };
+  'all reads normalized': {
+    year: number;
+  };
+  'refereed reads': {
+    year: number;
+  };
+  'refereed reads normalized': {
+    year: number;
+  };
+}
+
+export interface IADSApiMetricsResponse {
+  'basic stats': {
+    'average number of downloads': number;
+    'average number of reads': number;
+    'median number of downloads': number;
+    'median number of reads': number;
+    'normalized paper count': number;
+    'number of papers': number;
+    'recent number of downloads': number;
+    'recent number of reads': number;
+    'total number of downloads': number;
+    'total number of reads': number;
+  };
+  'basic stats refereed': {
+    'average number of downloads': number;
+    'average number of reads': number;
+    'median number of downloads': number;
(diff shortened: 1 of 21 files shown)


src/api/lib/metrics/types.ts
```diff
@@ -0,0 +1,104 @@
+export interface IADSApiMetricsParams {
+  bibcode: string;
+}
+
+export interface ICitationsHistogram {
```
thostetler, line 5: I think you could simplify this a bit by switching the type to be a union on these strings

```
type CitationsHistogramKey = 'nonrefereed to nonrefereed' | 'nonrefereed to nonrefereed normalized' | ...;

export type CitationsHistogram = Record<CitationsHistogramKey, { year: number }>
```

Something like that anyway

shinyichen: Updated

thostetler:
So a couple of things with this I think we should do differently.

1. Create a Metrics component to hold all this logic currently in the `pages/metrics` 
2. Move the `graphUtils` file to be next to the metrics component
3. I'm not a huge fan of these really long string literal property names being passed around.  I think we should switch to using enums.  I made some in the types and pushed a commit, but I think it would be good to refactor this at some point.  It's okay for now.

src/utils.ts
```diff
@@ -52,3 +53,27 @@ export const getDocument = async (api: AdsApi, id: string): Promise<IOriginalDoc
...
 };
+
+export const getHasGraphics = async (api: AdsApi, bibcode: string): Promise<boolean> => {
+  const result = await api.graphics.query({
+    bibcode: bibcode,
+  });
+  return result.isErr() ? false : true;
+};
+
+export const getHasMetrics = async (api: AdsApi, bibcode: string): Promise<boolean> => {
```
thostetler, line 64: I'm thinking that these two helpers should maybe be moved into their respective api services instead.  Something like:

`api.metrics.hasMetrics(bibcode: string): Promise<boolean>`

src/pages/abs/[id]/metrics.tsx
```diff
@@ -1,8 +1,139 @@
...
+        )}
+      </article>
+    </AbsLayout>
+  );
 };
 
-export default BasicPage;
+export default MetricsPage;
+
+export const getServerSideProps: GetServerSideProps<IMetricsPageProps> = async (ctx) => {
```
Comment (thostetler) [inline src/pages/abs/[id]/metrics.tsx:56]: So I think we may want to try to simplify the serversideProps part here.  
We really only need the bibcode available on the page, then pass that as a prop to the `<Metrics>` component.  From there we should be doing all this extra stuff, that makes sure we don't have these interdependencies.

So I would say let's move all the metrics calls here to inside the component, we could make a custom hook or whatever, but that way we keep things encapsulated a little better

Comment (thostetler) [reply src/pages/abs/[id]/metrics.tsx:56]: The server is going to render the `<Metrics />` component, so it'll make all these calls anyway, but we can keep the page simpler

src/components/Metrics/graphUtils.ts
```diff
@@ -0,0 +1,265 @@
...
+  values: IPair[];
+}
+
+export interface IPair {
+  x: string;
+  y: number;
+}
+export interface ICitationTableInput {
+  refereed: {
+    [CitationsStatsKey.NORMALIZED_NUMBER_OF_CITATIONS]: number;
```
thostetler, line 16: You should be able to do:
```
refereed: {
  [key in CitationsStatsKey]: number;
}
```
Unless I'm missing this is using any different keys?

thostetler (approved):
I have a couple things I think we want to work on with this.  
I'm fine with letting this one merge though for now and refactoring after since we want to keep moving on these changes, please let me know if you want to keep updating, I can help

src/utils.ts
shinyichen: Done

PR #75, opened 2021-10-20
Work without JavaScript

shinyichen:
* A first attempt at making things work without JavaScript. Will be improving some of the UI later.
* Also included metadata since that commit fixes an error previously missed.

### DIFF
--- a/src/components/Metatags/Metatags.tsx
+++ b/src/components/Metatags/Metatags.tsx
@@ -1,7 +1,8 @@
 import { IDocsEntity } from '@api';
-import React from 'react';
-import Head from 'next/head';
 import { Esources } from '@api/lib/search/types';
+import { format } from 'date-fns';
+import Head from 'next/head';
+import React from 'react';
 
 const ADS_BASE_URL = 'https://ui.adsabs.harvard.edu';
 
@@ -42,9 +43,11 @@ export const Metatags = (props: IMetatagsProps): React.ReactElement => {
   }
 
   const getFomattedNumericPubdate = (pubdate: string) => {
-    const regex = /^(\d{4})-(\d{2})-(\d{2})$/;
-    const res = regex.exec(pubdate);
-    return `${res[1]}/${res[2]}`;
+    try {
+      return format(new Date(pubdate), 'yyyy/MM');
+    } catch (e) {
+      return '';
+    }
   };
 
   const getLastPage = (page_range: string) => {


src/components/Metatags/Metatags.tsx
```diff
@@ -0,0 +1,187 @@
...
+  const { doc } = props;
+
+  if (!doc) {
+    return null;
+  }
+
+  const getFomattedNumericPubdate = (pubdate: string) => {
+    const regex = /(....)-(..)-(..)/;
+    const res = regex.exec(pubdate);
+    return `${res[1]}/${res[2]}`;
```
thostetler, line 47: I think we should change this up a little bit, one you could make this regex more explicit somewhat by doing something like:
`/^(\d{4})-(\d{2})-(\d{2})$/`
But I think it would be better to use the native Date for this:

```
import { format } from 'date-fns';

try {
  return format(new Date(pubdate), 'yyyy/MM');
} catch (e) {
  // date breaks, return an empty string
  return '';
}
```

src/components/Metatags/Metatags.tsx
```diff
@@ -0,0 +1,187 @@
...
+  };
+
+  const getArXiv = (identifier: string[]) => {
+    identifier.forEach((id) => {
+      if (id.startsWith('arXiv')) {
+        return id;
+      }
+    });
+    return '';
+  };
```
thostetler, line 62: Not sure about this function, `forEach` won't return anything, assuming you're looking for `filter`?

src/components/Metatags/Metatags.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+      <meta name="twitter:title" content={title} />
+
+      <meta name="twitter:site" content="@adsabs" />
+
+      <meta name="twitter:domain" content="NASA/ADS" />
+
+      <meta name="twitter:image:src" content={logo} />
+
+      <meta name="twitter:creator" content="@adsabs" />
```
thostetler, line 184: I think these are fine for now, this is taken basically straight from how bumblebee is doing things.  
However I think we can modularize this a little bit in a future refactor so we aren't mixing so much logic into the jsx. 
I'm sure we can at least cut down on some of the redundancy anyway

src/pages/abs/[id]/abstract.tsx
```diff
@@ -115,53 +115,55 @@ const AbstractPage: NextPage<IAbstractPageProps> = (props: IAbstractPageProps) =
...
-                  </a>
-                ) : null}
-                {!isBrowser() && doc.author.length > showNumAuthors ? (
-                  <span>{` and ${doc.author.length - showNumAuthors} more`}</span>
-                ) : null}
-              </div>
+                        {'  '}
+                        {orcid && (
+                          <Link
+                            href={`/search?q=${encodeURIComponent(`orcid:${orcid}`)}&sort=${encodeURIComponent(
```
Comment (thostetler) [inline src/pages/abs/[id]/abstract.tsx:141]: use the `next/link` query config here

PR #93, opened 2021-11-02
Not implemented page and environment var for ADS base URL

shinyichen:
* Created a not implemented page and redirect login and register to it
* Put ADS base url in env var

### DIFF
--- a/.env.local
+++ b/.env.local
@@ -1,5 +1,4 @@
 API_HOST=https://devapi.adsabs.harvard.edu/v1
 PORT=8000
 COOKIE_SECRET=G7Kfbufs9QkrnLRFPPGnciaLY3JLH3r9BL6nAKNdRPEzLAA5wcpmT6gF8hMVjY9n
-NEXT_PUBLIC_ADS_BASE_URL=https://ui.adsabs.harvard.edu
-NEXT_PUBLIC_ADS_ABS_URL=https://ui.adsabs.harvard.edu/abs
+NEXT_PUBLIC_BASE_URL=https://ui.adsabs.harvard.edu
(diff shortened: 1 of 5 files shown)


.env.local
```diff
@@ -1,3 +1,5 @@
 API_HOST=https://devapi.adsabs.harvard.edu/v1
 PORT=8000
-COOKIE_SECRET=G7Kfbufs9QkrnLRFPPGnciaLY3JLH3r9BL6nAKNdRPEzLAA5wcpmT6gF8hMVjY9n
\ No newline at end of file
+COOKIE_SECRET=G7Kfbufs9QkrnLRFPPGnciaLY3JLH3r9BL6nAKNdRPEzLAA5wcpmT6gF8hMVjY9n
+NEXT_PUBLIC_ADS_BASE_URL=https://ui.adsabs.harvard.edu
+NEXT_PUBLIC_ADS_ABS_URL=https://ui.adsabs.harvard.edu/abs
```
marblestation, line 5: It is enough to just have one environment variable for the base URL, we do not need a second one for the abstract URL because the `/abs` part does not change and it can just be constructed from the base url. So I suggest we get rid of `NEXT_PUBLIC_ADS_ABS_URL` and we build it where it is necessary using the base URL + `/abs`.

For `NEXT_PUBLIC_ADS_BASE_URL` I also suggest to rename it. Let's avoid ADS and maybe use something like `PUBLIC_BASE_URL`.

marblestation (changes requested):
Thanks @shinyichen! I have a couple of suggestions

.env.local
shinyichen: Modified as suggested. The env var name is required to start with `NEXT_PUBLIC_` to allow client side to access it.

marblestation (approved):
Great! Nothing else from my end! thank you!!

PR #94, opened 2021-11-02
Add citation count to result number display

shinyichen:
Add citation count and normalized citation count to total result display.
* Updated search query to include stats
* Updated search query response to include the stats
Screen Shot 2021-11-02 at 10 48 13 AM
Screen Shot 2021-11-02 at 10 48 25 AM

### DIFF
--- a/src/api/lib/search/search.ts
+++ b/src/api/lib/search/search.ts
@@ -1,4 +1,3 @@
-import { SolrField } from '@api';
 import { AxiosError, AxiosRequestConfig } from 'axios';
 import { err, ok, Result } from 'neverthrow';
 import { ApiTargets } from '../models';
@@ -20,7 +19,7 @@ export class SearchService extends Service {
     };
   }
 
-  async query(rawParams: IADSApiSearchParams): Promise<Result<IADSApiSearchResponse['response'], Error | AxiosError>> {
+  async query(rawParams: IADSApiSearchParams): Promise<Result<IADSApiSearchResponse, Error | AxiosError>> {
     const params = this.normalizeParams(rawParams);
     const config: AxiosRequestConfig = {
       method: 'get',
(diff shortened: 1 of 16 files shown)


src/api/lib/search/search.ts
```diff
@@ -32,7 +31,9 @@ export class SearchService extends Service {
       this.request<IADSApiSearchResponse>(config).then(
         (result) => {
           result.match(
-            ({ response }) => resolve(ok(response)),
+            ({ response, stats }) => {
+              stats ? resolve(ok({ response, stats })) : resolve(ok({ response }));
+            },
```
thostetler, line 36: I'm not a big fan of this, since we have an extra property to deal with even though like 90% of the queries won't have stats, but I don't think we should create a different request or anything for this.
```suggestion
            ({ response, stats }) => resolve(ok({ response, stats }));
```
Should just always return `stats` now even if it's `undefined` most of the time, so it's consistent

thostetler (changes requested):

src/pages/search/index.tsx
```diff
@@ -57,10 +60,24 @@ const Form = (props: IFormProps): ReactElement => {
...
+      : null;
+  });
+
+  const normCitationCount = useSelector(searchService, (state) => {
+    const userSort = state.context.params.sort[0].split(' ')[0];
+    return userSort === 'citation_count_norm' && state.context.result.stats?.stats_fields?.citation_count_norm?.sum
+      ? truncateDecimal(state.context.result.stats.stats_fields.citation_count_norm.sum, 1)
+      : null;
+  });
+
```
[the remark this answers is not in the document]
shinyichen: I moved this part into NumFound component

src/api/lib/search/search.ts
shinyichen: Done.

PR #96, opened 2021-11-03
downsized and cropped background images

shinyichen:


marblestation (changes requested):
I have investigated a bit how nextjs works with images and it reassures me that it has a [built-in cache](https://nextjs.org/docs/api-reference/next/image#caching-behavior), so files are stored in a directory the first time they are generated and not recreated at each request. There are some parameters we can control like for how long they will be valid. Also, if necessary, we can set the [unoptimized parameter to true](https://nextjs.org/docs/api-reference/next/image#unoptimized) and then the quality, size or format won't be changed.

NextJS's default image loader [has built-in capabilities to change the size and the quality of the image](https://nextjs.org/docs/api-reference/next/image) (see [width 1920](https://playground.adsabs.harvard.edu/_next/image?url=%2F_next%2Fstatic%2Fimage%2Fpublic%2Fimg%2Fstar-bg-cropped.6a2cfe3044f4dd64ef8f9cb39cb67b8f.png&w=1920&q=75) vs [width 256](https://playground.adsabs.harvard.edu/_next/image?url=%2F_next%2Fstatic%2Fimage%2Fpublic%2Fimg%2Fstar-bg-cropped.6a2cfe3044f4dd64ef8f9cb39cb67b8f.png&w=256&q=75), or see [quality 75](https://playground.adsabs.harvard.edu/_next/image?url=%2F_next%2Fstatic%2Fimage%2Fpublic%2Fimg%2Fbg-bio.8920f22f764dd338eabdae83a6e09d03.jpg&w=1920&q=75) vs [quality 25](https://playground.adsabs.harvard.edu/_next/image?url=%2F_next%2Fstatic%2Fimage%2Fpublic%2Fimg%2Fbg-bio.8920f22f764dd338eabdae83a6e09d03.jpg&w=1920&q=25)). Unfortunately, it does not allow cropping by default so the cropping you have done is great!

Also, it looks like it can change the quality of JPEG images but it does nothing with PNG images. So, could we create a JPEG version of the file `star-bg-cropped.png` and name it following the same pattern as the rest (i.e., `bg-astro.jpg`)?

BTW, side note, it also has the [placeholder attribute](https://nextjs.org/docs/api-reference/next/image#placeholder) that might be interesting for our purpose of showing something while the image is loading. It might be worth exploring.

shinyichen:
@marblestation I reverted back the image resolution and just cropped them. Also took the jpg version of the starry background, renamed it and cropped it.

The placeholder defaults to empty, which means show nothing, so nothing needs to be done there.

marblestation (approved):
It looks good! Let's try this out in real life

PR #99, opened 2021-11-03
Compact layout

shinyichen:
* Make display of results and abstract text more compact.
* Subpage parsing fix

Screen Shot 2021-11-03 at 1 22 36 PM

Screen Shot 2021-11-03 at 1 22 57 PM

### DIFF
--- a/src/components/AbstractSideNav/AbstractSideNav.tsx
+++ b/src/components/AbstractSideNav/AbstractSideNav.tsx
@@ -5,6 +5,7 @@ import { DocumentIcon } from '@heroicons/react/outline';
 import { ChevronDownIcon } from '@heroicons/react/solid';
 import { useViewport, Viewport } from '@hooks';
 import { useAppCtx } from '@store';
+import { useBaseRouterPath } from '@utils';
 import clsx from 'clsx';
 import Link from 'next/link';
 import { useRouter } from 'next/router';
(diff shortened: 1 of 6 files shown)


src/components/ResultList/Item/Item.tsx
```diff
@@ -84,30 +84,29 @@ export const Item = (props: IItemProps): ReactElement => {
...
+            <div className="text-sm">
               {author.slice(0, 10).join('; ')}
               {author_count > 10 && <span className="text-xs italic"> and {author_count - 10} more</span>}
             </div>
           )}
           <div className="flex py-1">
             <span className="text-xs">
               {formattedPubDate}
-              {formattedPubDate && formattedBibstem ? ' | ' : ''}
+              {formattedPubDate && formattedBibstem ? <>&nbsp;&nbsp;·&nbsp;&nbsp;</> : ''}
```
thostetler, line 105: maybe dumb question by is this better than doing:
```suggestion
              {formattedPubDate && formattedBibstem ? <span className="px-1">·</span> : ''}
```
or whatever padding would look good

src/components/AbstractSideNav/AbstractSideNav.tsx
```diff
@@ -19,7 +19,7 @@ export interface IAbstractSideNavProps extends HTMLAttributes<HTMLDivElement> {
 
 export const AbstractSideNav = ({ doc }: IAbstractSideNavProps): ReactElement => {
   const router = useRouter();
-  const subPage = last(router.asPath.split('/'));
+  const subPage = last(router.asPath.split('/')).split('?')[0];
```
thostetler, line 22: There is a utility for splitting on `?`
```suggestion
  const basePath = useBaseRouterPath();
  const subPage = last(basePath);
```

thostetler (changes requested):
LGTM - couple small changes

marblestation (approved):
I am not checking the implementation, only the look & feel, and I like it!

src/components/ResultList/Item/Item.tsx
shinyichen: You are right!

shinyichen: Changed

PR #101, opened 2021-11-04
Fixes pagination history issue

thostetler:
Moves Pagination into full component
Improvements to `usePagination` hook
Adds tests for `usePagination` hook
Gets storybook working properly for pagination


src/utils.ts
```diff
@@ -66,7 +65,27 @@ export const getFomattedNumericPubdate = (pubdate: string): string | null => {
...
+ * extract the current page number from the query
+ */
+export const parsePageFromQuery = (query: ParsedUrlQuery): number => {
+  try {
+    const { p } = query;
+    const page = parseInt(Array.isArray(p) ? p[0] : p, 10);
+    if (page < 1) {
+      return 1;
+    }
+    return page === 0 || Number.isNaN(page) ? 1 : page;
```
shinyichen, line 87: The check `isNan(page)` comes after `if (page < 1)` comparison?

thostetler: Yep you're right, I'll fix!

PR #111, opened 2021-11-10
Implement volume content section

shinyichen:
* An example doc for testing: 1990ARA&A..28..215D
Screen Shot 2021-11-10 at 3 04 20 PM


marblestation (approved):
This looks good to me!

PR #112, opened 2021-11-12
Check if token is expired before bootstrap

thostetler:
- Bootstrapping is now completely controlled by the api service. 
- Token is injected into all api requests and persisted via localStorage
- On 401 errors, token is refreshed and the request is retried
- Exposes `bootstrap` endpoint as a normal service, rather than a static method
- Moves token management to api `utils`


marblestation:
Quick note: Not sure if this PR is ready to be reviewed, but FYI, in its current state it does not build

```Failed to compile.

./src/store/queries.ts:7:22
Type error: Property 'isValid' does not exist on type 'typeof Adsapi'.

   5 |   useQuery<IUserData>(['bootstrap'], fetchBootstrap, {
   6 |     onSuccess,
>  7 |     enabled: !Adsapi.isValid(userData),
     |                      ^
   8 |   });
   9 | };
  10 |
error Command failed with exit code 1.
```

thostetler:
This file should have been deleted, I'll add the commit in

marblestation (approved):
This is now working as expected! Thanks!

PR #122, opened 2022-01-31
Clear selected docs on search

thostetler:
Selected docs should now clear on search, but not when changing pages.

Depends on #121


thostetler:
Depends on #121

shinyichen (dismissed):

shinyichen (approved):

PR #124, opened 2022-02-01
UI refactor landing

shinyichen:
Landing page, classic form and paper form

Depends on #123


thostetler:
this LGTM, I added a couple commits to address some things -- I would be careful replacing native elements `<a>` `<button>` with generic ones like `<Box>` since we risk losing some semantic information. 
In this case, it was easy enough to just do `as={'a'}` or whatever, so no biggie.  But as it was, it wasn't accessible.

PR #126, opened 2022-02-01
UI refactor - abstract

shinyichen:
Depends on #125


src/components/Select/Select.tsx
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export interface ISelectProps {
+  formLabel?: string;
+  options: SelectOption[];
+  defaultOption: string;
+  onOptionSelected: (id: string) => void;
+}
+
+export const Select = ({ formLabel, options, defaultOption, onOptionSelected }: ISelectProps): ReactElement => {
```
thostetler, line 20: Why use this custom component instead of Chakra's built-in `Select`? https://chakra-ui.com/docs/form/select

thostetler (approved):
This LGTM, other than a comment I had.

I also pushed a small commit, I memoized a couple things.  possibly premature optimization though...

src/components/Select/Select.tsx
shinyichen: Chakra's select dropdown is native. Here I am using Chakra's menu as dropdown with a customized label.

shinyichen: Looks like in the next PR, I updated this component to use ReactSelect rather than Chakra's menu.

PR #136, opened 2022-02-03
Updates to pages and components

thostetler:
Depends on #134


src/components/ResultList/Item/Item.tsx
```diff
@@ -66,20 +52,11 @@ export const Item = (props: IItemProps): ReactElement => {
 
   return (
     <Flex direction="row" as="article" border="1px" borderColor="gray.50" mb={1} borderRadius="md">
-      <Flex direction="row" className={checkBgClass} m={0}>
-        <Text color={state.matches('selected') ? 'white' : 'initial'} display={{ base: 'none', md: 'initial' }} mr={1}>
-          {index}
+      <Flex direction="row" m={0}>
```
shinyichen, line 55: I believe here the style was accidentally removed. 
```suggestion
      <Flex
        direction="row"
        backgroundColor={isChecked ? 'blue.500' : 'gray.50'}
        justifyContent="center"
        alignItems="center"
        mr="2"
        px="2"
        borderLeftRadius="md"
      >
```

shinyichen (commented):
Not sure if this is all the changes from your branch or there will be more that might already fixed these. I'm getting a couple of errors in the server side console log.
1) component/export/hooks.ts still references `useAppCtx`, and also in `useSearchMachine.ts` (although I don't think this module is used anymore)
2) Error  with the `AbstractSideNav`:
```
TypeError: Cannot read property 'undefined' of undefined
    at fetchHasMetrics (webpack-internal:///./src/components/AbstractSideNav/queries.ts:53:129)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

thostetler:
@shinyichen yep, still lots to come in -- with this PR though you should be able to get to the search page without error -- working on the rest tonight

shinyichen:
> @shinyichen yep, still lots to come in -- with this PR though you should be able to get to the search page without error -- working on the rest tonight

I have no other comments then. Looks good.

PR #143, opened 2022-02-19
Upgrade Next to 12.1.0

thostetler:
Also a couple small build updates

PR #145, opened 2022-02-23
Fixes to jest and storybook

thostetler:
Update storybook to use webpack 5 
Fix path aliases in both storybook and jest 
Add global decorator to work with new store provider 
Fix global type conflicts with jest and cypress 
Update searchbar test as example of how to use `composeStories`

depends on: #143


src/store/store.ts
```diff
@@ -2,53 +2,49 @@ import { IUserData } from '@api';
...
+  if (process.env.NODE_ENV === 'test') {
+    return create<AppState>(state);
+  }
+
   const store = create<
     AppState,
     SetState<AppState>,
     GetState<AppState>,
-    StoreApiWithPersist<AppState> & StoreApiWithDevtools<AppState>
+    Mutate<StoreApi<AppState>, [['zustand/persist', Partial<AppState>], ['zustand/devtools', never]]>
```
shinyichen, line 32: @thostetler Can you help me understand what these are?

thostetler: Hey @shinyichen, yeah this was a suggested deprecation in the types when I was refactoring
I just looked at https://github.com/pmndrs/zustand/blob/0fce3a9e7460002cf66222c86138095a47474121/tests/middlewareTypes.test.tsx#L380

to find out how to handle these types.

This is basically so that the middlewares (`devtools` and `persist`) are properly typed.  This is a cleaner way of doing things, so much preferred!

shinyichen (approved):

PR #149, opened 2022-03-03
Fix loading bugs

thostetler:
Issue after minifying code with SWC minifier

Turns that off, and also upgrade react-query


shinyichen (approved):
Cool, works now.

PR #153, opened 2022-03-10
Some small fixes

thostetler:
- move focus to input on search example selection
- center abstract preview loading spinner


shinyichen (approved):
This looks good to me. I just found a bug with the search examples, wonder if it's a small fix or not. 

When I open the landing page (my preference is Astrophysics), and click on the examples, such as `author:"huchra, john"`, what goes to the search bar is `author:"penrose, roger"`. But if I switch to a different theme and come back to Astrophysics, then click on `author:"huchra, john"` and this time what goes to the search bar is correct.

Reload the page again, and the same thing starts over.

PR #156, opened 2022-03-14
Adds ListActions back in

thostetler:
- Updates to use store
- Gets select all/none working properly
- Refactor Sort component, get it working properly

- Adds Why did you render utility for debugging performance issues


src/store/slices/docs.ts
```diff
@@ -61,5 +82,30 @@ export const docsSlice: StoreSlice<IAppStateDocsSlice> = (set, get) => ({
...
+      }),
+      false,
+      'docs/selectAll',
+    ),
+
+  // remove items that exist in both lists, but keep the others in `selected` intact
+  clearAllSelected: () =>
+    set(
+      (state) => ({
+        docs: { ...state.docs, selected: difference(state.docs.selected, state.docs.current), isAllSelected: false },
```
shinyichen, line 103: I don't really understand the purpose of the function here. Is it suppose to clear all so nothing is selected? So `selected` should be empty?

shinyichen (commented):
Just an observation, maybe it belongs to a separate PR. While using class form, if I change the sort, the results page is still sorted by date, although the URL shows the right sort I selected. Maybe related, the handleOnSubmit in the cassic-form is not being used.

src/store/slices/docs.ts
thostetler: So I should have thought up a better name/way to do it.

But the problem is that we want an overall selection that persists as you page through results, but have the `Select All` functionality work page to page.  So it should clear after each pagination.  So we want to only clear the docs that exist in both lists, but leave the rest of `selected` intact.
So in a way, there shouldn't be a time when we're completely clearing the selected docs (unless we are doing a new search, for example)

src/components/ResultList/ListActions.tsx
```diff
@@ -1,7 +1,93 @@
...
+  const selectAll = useStore((state) => state.selectAll);
+  const clearAllSelected = useStore((state) => state.clearAllSelected);
+  const isAllSelected = useStore((state) => state.docs.isAllSelected);
+
+  return (
+    <>
+      <Button variant="link" fontWeight="normal" disabled={isAllSelected} onClick={selectAll}>
+        Select All
+      </Button>
+      <Button variant="link" fontWeight="normal" disabled={!isAllSelected} onClick={clearAllSelected}>
```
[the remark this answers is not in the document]
shinyichen: Select none is only enabled when all is selected. But should be enabled if anything on the page is selected. 
(In the mean time, I will be thinking about the UI for clearing all selections)
Screen Shot 2022-03-22 at 11 04 14 AM

PR #167, opened 2022-03-23
Updated UI to show and handle selection across result pages

shinyichen:
A little change on the UI  so that selection across multiple results pages make more sense.
https://user-images.githubusercontent.com/636361/159738006-063556c8-b891-4c8d-bc8c-f1a7aabbafea.mov


thostetler (approved):
Looks pretty good to me,
we should probably come up with better names at some point since it gets a little confusing looking at the store.  But the changes look good!

PR #168, opened 2022-03-23
Pagination and history fixes

thostetler:


shinyichen:
I am doing the following test:
1. Do a search
2. Go to page 2
3. Do a second search --> url still shows p=2, but content is page 1 of second search
4. Use back button to go back --> url goes back to first search, p=1, but content is still from second search

PR #169, opened 2022-03-23
Search when input is empty

shinyichen:
Change how to handle empty search. Rather than disable search, check if query is empty on submit, don't handle submit if it is empty. This way we don't disable search in JS disabled browsers.

Also use loading indicator on submitting search on landing page.


thostetler:
LGTM assuming ready to review, please merge when ready

PR #172, opened 2022-03-25
Couple fixes

thostetler:
Adds params for prefetchers for metrics/graphics
Updates path for side nav

Adds path rewrite for link_gateway, so those don't resolve to 404

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -9,6 +9,16 @@ const config = {
   distDir: process.env.DIST_DIR || 'dist',
   poweredByHeader: false,
   reactStrictMode: true,
+  async rewrites() {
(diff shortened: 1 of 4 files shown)


next.config.js
```diff
@@ -9,6 +9,15 @@ const config = {
   distDir: process.env.DIST_DIR || 'dist',
   poweredByHeader: false,
   reactStrictMode: true,
+  async rewrites() {
```
shinyichen, line 12: I am getting this error with `yarn build`
`Error: rewrites must return an array, received undefined.`

shinyichen (approved):
No other comments besides the build error.

PR #180, opened 2022-04-12
Explorer metrics

shinyichen:
This is the initial work on the explorer pages. Right now only the metrics page is there. 
## Some implementation details
* If nothing from search results is selected, use the original search query (part of the url) to get the list of bibcodes for the viz. If selections made, use the vault to generate an id (pass qid param in the url), the qid is used to get the list of bibcodes for the viz.
* The original search query from the url is also how we get 'back to the search results' from viz.
* Use `useQueries` to make parallel queries to get bibcodes (1000 rows) at a time. Limited to 7000 bibcodes for metrics.
* Only after all the bibcodes have been fetched, we make fetch metrics query.
* If more than 6000 bibcodes, don't get the timeseries (used for indices graph). It is fetched separately.
* During SSR, prefetch the list of bibcodes first. Metrics data are fetch on the client side.

## Things to be done
* Refactor out the search bar section so it can be shown above the visualization tabs
* All other tabs

Screen Shot 2022-04-12 at 2 42 18 PM


shinyichen:
@thostetler This PR is ready for review. I just push the last commit, which I tried to move some of the data fetching to server side. Please see if that's appropriate.

shinyichen:
@thostetler It's ready for review.

thostetler (approved):
@shinyichen I added a few commits. Basically:

- Removed `search/bigquery` since we have this functionality in `vault` and I don't recall it being used anywhere in bbb directly like that.

- Made some modifications to the `parseQueryFromURL` utility, removing the one you added.  I think it could be made more general by adding a generic to allow for additional typed properties.

- In places where you were using `refetch` and `enabled: false` to trigger the fetches, I refactored quick. In the case that you're watching some conditional for when to fetch, you can pass it directly to `enabled`. This way it doesn't **force** a fetch from the server, and will look in the cache.

- Changed the sort postfix back to `bibcode desc` since that's what's used on BBB. But updated the pass in the tiebreaker sort when creating params for the infinite query.

- Removed the page props for `numFound` and `query` since we can derive them from async state.

- Added a prefetch for infinite query, however for now this only does the first "page" which is 1000 records. Have to keep playing with that, I have some ideas.

So all in all not a ton of changes, but some tweaks to improve how we're handling the queries and typings. Please check them out and let me know, we can move them to a separate PR if you'd like to make more changes so that the bulk of your work makes it in. 
Nice work, the interface looks pretty good!

PR #185, opened 2022-04-20
Don't show toast message if fetch error is due to has no graphics or has no metrics

shinyichen:
In abstract page, we get toast message if fetching getHasMetrics and getHasGraphics return negative. Here I added a check to see if the error messages match no graphics or no metrics error. I don't really like doing it this way. @thostetler do you have any suggestions?


thostetler:
@shinyichen I added a commit with a little different way of doing it (https://github.com/adsabs/nectar/pull/185/commits/04db47148e73ec156ca2b2d292723394f345df4b)

Now we can add `skipGlobalErrorHandler: true` to the `meta` of any request and it'll skip the global one.  That way the logic to skip stays close to the query and it's easier to differentiate between query keys.

thostetler (approved):

PR #186, opened 2022-04-20
Integration tests for landing page

shinyichen:
Tests on:
* Search bar behaviors
* Theme selections
* Search examples

### DIFF
--- a/cypress/integration/landing.spec.ts
+++ b/cypress/integration/landing.spec.ts
@@ -1,0 +1,117 @@
+import { examples } from '@components/SearchExamples/examples';
+import { themes } from '@components/NavBar/models';
+
+const baseUrl = Cypress.config().baseUrl;
+
+const themeOptions = Object.values(themes);
+
+describe('Landing Page', () => {
+  beforeEach(() => {
+    cy.visit('/');
+    cy.get('form').find('input[type="text"]', { timeout: 10000 }).first().as('input').should('be.enabled');
+  });
+  describe('Search bar behaviors', () => {
+    it('searchbar clear should clear input', () => {
+      cy.getByTestId('searchbar-clear').should('not.exist');
+      cy.get('@input').type('star');
+      cy.get('@input').should('have.value', 'star');
+      cy.getByTestId('searchbar-clear').click();
+      cy.get('@input').should('have.value', '');
+      cy.getByTestId('searchbar-clear').should('not.exist');
+    });
+
+    it('searchbar should not search empty term', () => {
+      cy.get('@input').type(`{enter}`);
+      cy.url().should('eq', baseUrl + '/');
+      cy.getByTestId('searchbar-submit').click();
+      cy.url().should('eq', baseUrl + '/');
+    });
+
+    it('search using enter should go to results page', () => {
+      cy.get('@input').type(`star{enter}`);
+      cy.url().should('eq', baseUrl + '/search?q=star&sort=date+desc&p=1');
+    });
+
+    it('search using submit button should go to results page', () => {
+      cy.get('@input').type('star');
+      cy.getByTestId('searchbar-submit').click();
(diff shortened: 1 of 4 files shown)


cypress/integration/landing.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+
+const baseUrl = Cypress.config().baseUrl;
+
+const themeOptions = Object.values(themes);
+
+describe('Landing Page', () => {
+  beforeEach(() => {
+    cy.visit('/');
+    cy.get('form').find('input[type="text"]', { timeout: 10000 }).first().as('input').should('be.enabled');
+  });
```
thostetler, line 12: This is okay, although I would say that we may wanna not do this in a unit test style, since it increases the run time significantly.
Probably we could improve on the `results.spec` file in this way too.
If we can structure them so that we only do a `cy.visit` a minimum amount of times that would be ideal.

cypress/integration/landing.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+      cy.getByTestId('searchbar-submit').click();
+      cy.url().should('eq', baseUrl + '/search?q=star&sort=date+desc&p=1');
+    });
+  });
+
+  describe('Themes selection', () => {
+    it('All themes should be available and selection should make theme change', () => {
+      themeOptions.forEach((themeOption, i) => {
+        cy.get('#theme-selector').click();
+        cy.get(`#react-select-theme-selector-option-${i}`).should('have.text', themeOption.label);
```
thostetler, line 46: Is this numbered option how `react-select` does it's IDs? 
seems a little fragile since it relies heavily on the order of the themes.  If we change the name it breaks (that's correct), but also changing the order, or adding any new ones (other than the end).
Probably would be better to replace with `...option-astrophysics`, `...option-general`, etc.

cypress/integration/landing.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+      cy.get('a[href="/classic-form"]').should('not.exist');
+      cy.get('a[href="/paper-form"]').should('not.exist');
+      cy.get('#theme-selector').click();
+      cy.get('#react-select-theme-selector-option-4').click();
+      cy.get('a[href="/classic-form"]').should('not.exist');
+      cy.get('a[href="/paper-form"]').should('not.exist');
+      cy.get('#theme-selector').click();
+      cy.get('#react-select-theme-selector-option-5').click();
+      cy.get('a[href="/classic-form"]').should('not.exist');
+      cy.get('a[href="/paper-form"]').should('not.exist');
```
thostetler, line 97: Same as above, if the ordering were to change or something it would break.  Maybe that's okay, but in this test we're only really concerned with whether classic/paper forms disappear when `theme !== 'astrophysics'`.
I think we could simplify and make this test more resistant to change

thostetler (approved):
Looks pretty good, I have some general structural concerns that I mention, otherwise nice work on these tests!

cypress/integration/landing.spec.ts
shinyichen: I was also expecting (at least based on `react-select` documentation) that the ID naming would be based on the option's value and not a number. But turns out it's not. I looked at the DOM on `react-select` element on their official site and the ID naming is also based on numbers. 

Since we are using the 'themes' from models, the tests should be ok if model changes, right?

cypress/integration/landing.spec.ts
shinyichen: I have made changes to make it more like integration tests.

cypress/integration/landing.spec.ts
shinyichen: These have been changed so that I am not making any assumptions about the theme ordering.

cypress/integration/landing.spec.ts
thostetler: Weird! 
Oh good point, I suppose it would synchronize in both spots.  Yep then this looks good 👍

PR #187, opened 2022-04-25
Consolidate API

thostetler:
Moves `_api` to `api`
Updates/fixes broken paths
Adds vault and reference endpoints 
Update to server-side api middleware

Paper form is temporarily broken since it's a larger refactor, will work on that in subsequent PR


thostetler:
Merging this since it's blocking and mainly path updates

PR #192, opened 2022-05-03
Quickfields

shinyichen:
## Add quick field and all search term dropdown
Screen Shot 2022-05-03 at 5 07 54 PM


thostetler (commented):
One potential problem is that `<QuickFields/>` being part of `SearchBar` causes it to be rendered every keystroke.

quick-fields-render

I think we can get most of the way there by moving `QuickFields` into it's own component directory.  Then use the store `updateQuery` to update.

Have to think about how to handle the cursor though.

thostetler:
Searchbar probably needs some refactoring to keep the main component separate from the store updates so we can have extraneous components included without extra re-renders

shinyichen:
@thostetler Maybe we can discuss how to refactor next. Since I wanted to also add the search component on the visualization pages (like metrics) too.

> Searchbar probably needs some refactoring to keep the main component separate from the store updates so we can have extraneous components included without extra re-renders.
> 
> For now, it's much improved just wrapping your callback with a `useCallback`

thostetler (approved):
This LGTM

PR #194, opened 2022-05-06
Refactor AbstractSideNav

thostetler:
This restructures a bit the way the menu is rendered.  Didn't change any styling, although I think there may be a few slight differences.
Mainly this was to cleanup and fix how the links are being handled so that dynamic routes work properly


shinyichen (approved):
Just a couple of small things I noticed, nothing else.

PR #199, opened 2022-05-12
All search terms dropdown

shinyichen:
Rewrote the All Search Terms Dropdown without other UI library, so we can have both filtering and tooltips
Screen Shot 2022-05-12 at 4 58 46 PM


thostetler:
LGTM!

PR #201, opened 2022-05-17
Remove duplicated fetch calls

shinyichen:


thostetler:
I believe we'll have to keep both, since they produce different entries in the cache.  So these aren't duplicated, just need to do an extra step if we've been passed a search rather than a `qid`

PR #209, opened 2022-06-03
Scix 129 bug classic form has wrong labels or ignores fields

shinyichen:
* Fix classic form property field 'articles only'
* Fix property field being ignored
* Fix sort default


thostetler (approved):
Looks good

PR #211, opened 2022-06-07
Scix 135 implement highlights and show the button again

shinyichen:


shinyichen:
@thostetler Just marking this PR as ready but I don't want to distract you from the more important tasks you are on. Just whenever you are ready to review.

src/pages/search/index.tsx
```diff
@@ -46,13 +48,40 @@ const SearchPage: NextPage = () => {
...
     updateQuery({ sort });
     void router.push({ pathname: router.pathname, search }, null, { scroll: false, shallow: true });
   };
 
+  const handleShowHighlights = (show: boolean) => {
+    if (show) {
+      setShowHighlights(true);
+    } else {
+      setShowHighlights(false);
+    }
```
thostetler, line 82:
```suggestion
    setShowHighlights(show)
```

src/pages/search/index.tsx
```diff
@@ -46,13 +48,40 @@ const SearchPage: NextPage = () => {
...
+
+  useEffect(() => {
+    if (!highlightsIsLoading && highlightsIsError) {
+      toast({
+        status: 'error',
+        title: 'Error!',
+        description: 'Unable to fetch highlights',
+      });
+    }
+  }, [highlightsIsLoading, highlightsIsError]);
```
thostetler, line 69: I think we should move all this highlights related stuff into a hook at least or maybe build it into the toggle component if possible.  We can use the cache to pass the data around if needed.

thostetler (changes requested):
This looks great, but I'd like to change how it works a bit. 
We should do all the highlights related fetching and restructuring in one spot.  I would suggest doing it when we toggle the button.

Then just getting the response from the cache where we need it.  I can help then with some more suggested changes.

src/components/ResultList/Item/Item.tsx
```diff
@@ -110,13 +124,34 @@ export const Item = (props: IItemProps): ReactElement => {
...
+const Highlights = ({ isLoading, highlights }: { isLoading: boolean; highlights: string[] }) => {
+  return (
+    <Box as="section" aria-label="highlights" className="search-snippets" my={2} data-testid="highlights-section">
+      {isLoading || !highlights ? (
+        <CircularProgress mt={5} isIndeterminate size="20px" />
+      ) : (
+        <>
+          {highlights.length > 0 ? (
+            highlights.map((hl, index) => (
+              <Text key={`hl-${index}`} dangerouslySetInnerHTML={{ __html: decode(hl) }}></Text>
```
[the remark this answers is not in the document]
shinyichen: The highlight text is encoded, like this one:
`the complete evolutionary track of a 1 M &lt;SUB&gt;sun&lt;&#x2F;SUB&gt; <em>star</em> from the pre-main sequence (PMS) to a cooling` 
and shows up as 
`the complete evolutionary track of a 1 M <SUB>sun</SUB> star from the pre-main sequence (PMS) to a cooling` in the browser.

shinyichen:
@thostetler I made some changes, let me know if it's what you meant or if there's anything else. Thanks.

thostetler:
@shinyichen, that's a start, but I think we can go further.
Take a look at my implementation on top of the work you did: https://github.com/adsabs/nectar/commit/73761ce0e4b8075770c5299dde6df24fda30cd25#diff-47ca9af98444343931ff79ad3246e31cf32c01d7a9580db6948e025e88c6b363

I removed all the references to highlights from top level components like `<SearchPage>` and from the general hook.  We should co-locate these components to where they are being used.

For example in my commit, I moved all the fetching into a hook that sits next to `<SimpleResultsPage/>` since it's not used elsewhere and we can contain potential changes to those files.  I moved the toggle state to the store, so that we can use context to check state, and not have pass so many props around.

The hook does all the work of fetching and transforming, then passes the data to `<Item/>`, similar to how you were doing it.  

Please check it out when you get a chance, I didn't want to push onto your branch yet.  I'm happy to discuss any of this then when you'd like!

thostetler (approved):

PR #226, opened 2022-07-01
Bug fixes

shinyichen:
Fix three minor issues:
* Specify zIndex for 'Select' component. Fix problem with the dropdown appearing under some other components.
* Subtract original query from 'Trending' operation
* Remove quotes from some quick search terms


src/components/ResultList/ListActions.tsx
```diff
@@ -348,3 +346,7 @@ const ExportMenu = (props: MenuGroupProps & { exploreAll: boolean }): ReactEleme
     </MenuGroup>
   );
 };
+
+const createOperatorQuery = (operator: Operator, originalQuery: string) => {
+  return operator === 'trending' ? `${operator}(${originalQuery})-(${originalQuery})` : `${operator}(${originalQuery})`;
```
thostetler, line 351:
```suggestion
  return operator === 'trending' ? `trending(${originalQuery})-(${originalQuery})` : `${operator}(${originalQuery})`;
```

src/components/ResultList/ListActions.tsx
```diff
@@ -94,7 +92,7 @@ export const ListActions = (props: IListActionsProps): ReactElement => {
   const handleOperationsLink = (operator: Operator) => {
     if (exploreAll) {
       // new search with operator
-      const q = `${operator}(${router.query.q as string})`;
+      const q = createOperatorQuery(operator, router.query.q as string);
```
thostetler, line 95:
```suggestion
      const query = parseQueryFromUrl(router.query);
      const q = createOperatorQuery(operator, query.q);
```
Just so that the type is already set and you don't have to assert.

thostetler (approved):
Looks good 👍 

Couple of small suggestions

PR #229, opened 2022-07-05
Get keyformat working for export

thostetler:
`keyformat` works now for export component
Now properly escapes/cleans input from `keyformat`

Adds a purify utility for doing this type of escaping, still a little wip since I wasn't able to get types to work perfectly


shinyichen (approved):
Just one question: The old code had a default keyformat `%R` which is not there after the update. Should that have been kept?

thostetler:
@shinyichen yep `%R` produces a normal bibcode, so it's the same as omitting the prop.  But maybe it would be best to keep the default... I'll update

PR #230, opened 2022-07-06
First pass at adding math rendering

thostetler:
Adds simple latex math rendering via MathJax.
Should this be server-rendered? if so, have to think about how best to handle it.

What other fields will need math rendering? it's overkill to run it on the whole page

Needs some test bibcodes that have tons of math in their abstracts/titles etc for testing


thostetler:
For reference, the BBB config looks like this:
```
{
  messageStyle: 'none',
  HTML: ['input/TeX', 'output/HTML-CSS'],
  TeX: {
    extensions: ['AMSmath.js', 'AMSsymbols.js'],
    equationNumbers: { autoNumber: 'AMS' },
  },
  extensions: ['tex2jax.js'],
  jax: ['input/TeX', 'output/HTML-CSS'],
  tex2jax: {
    inlineMath: [
      ['$', '$'],
      ['\\(', '\\)'],
    ],
    displayMath: [
      ['$$', '$$'],
      ['\\[', '\\]'],
    ],
    processEscapes: true,
  },
  'HTML-CSS': {
    availableFonts: ['TeX'],
    linebreaks: { automatic: true },
  },
}
```

shinyichen (approved):

PR #231, opened 2022-07-13
Scix 136 implement author network

shinyichen:
# Components
Screen Shot 2022-07-13 at 11 06 50 AM
# Details
Screen Shot 2022-07-13 at 11 20 25 AM


src/components/Visualizations/Containers/AuthorNetworkPageContainer.tsx
```diff
@@ -0,0 +1,393 @@
+import { IADSApiSearchParams, useSearch, useVaultBigQuerySearch } from '@api';
```
thostetler, line 1: I think we should probably break this file apart into several components.  They could all live inside this file, but I think that would make it simpler.

thostetler (approved):
I know this is still WIP.  I think the work looks pretty good -- my only comment would be to break some components up a little to improve readability and make simpler

shinyichen:
Starting from this commit -> [Author network graph using D3 (94ba441)](https://github.com/adsabs/nectar/pull/231/commits/94ba441cc024562bd7d97e5db8c56d076b7d9a8b) are the new commits for replacing with D3 library

src/components/Visualizations/Containers/AuthorNetworkPageContainer.tsx
shinyichen: @thostetler I've moved the new changes for D3 and the refactoring to #233 
This PR has nothing new since you last reviewed.
Let me know if you feel this PR can be merged

PR #235, opened 2022-08-08
Improvement on Author network

shinyichen:
* Use reducer to better manage state


thostetler:
This is looking better to me! I don't see anything else to update right now.  Thanks for making the changes!

thostetler:
This LGTM, thanks for the updates!

PR #242, opened 2022-09-28
Scix 197 graph data download

shinyichen:
Add the download link for all graph data.

Screen Shot 2022-09-28 at 3 17 38 PM


src/components/DataDownloader/DataDownloader.tsx
```diff
@@ -0,0 +1,57 @@
+import { FC, HTMLAttributes, useEffect, useRef, useState } from 'react';
```
thostetler, line 1: I have a hook I made for the export component that handles downloading files if it would be of use here

https://github.com/adsabs/nectar/blob/master/src/hooks/useDownloadFile.ts

src/components/Visualizations/GraphPanes/YearsGraphPane.tsx
```diff
@@ -31,6 +32,8 @@ export const YearsGraphPane = ({ data, onApplyCondition }: IYearsGraphPaneProps)
     }
   }, [data]);
 
+  console.log(baseGraph);
```
thostetler, line 35:
```suggestion
```

src/components/Visualizations/GraphPanes/YearsGraphPane.tsx
```diff
@@ -93,10 +96,27 @@ export const YearsGraphPane = ({ data, onApplyCondition }: IYearsGraphPaneProps)
...
 
+  const getCSVDataContent = () => {
+    let content = 'Year, Article Count, Ref Count';
+
+    baseGraph.data.forEach(({ year, refereed, notrefereed }) => {
+      content += `\n${year},${refereed + notrefereed},${refereed}`;
+    });
+
+    return content;
+  };
```
thostetler, line 107:
```suggestion
  const getCSVDataContent = into(
    'Year, Article Count, Ref Count',
    map(({ year, refereed, notrefereed }) => `\n${year},${refereed + notrefereed},${refereed}`))
  );
```

Haven't actually tested this out, but you should be able to do something like this if you wanted to use the utils.  (will need to also add `into` and `map` from ramda)

thostetler (changes requested):
Couple of things, but looks good

I would look at my hook to see if you can use inside your component.
Feel free to merge when you're ready then

PR #244, opened 2022-10-07
Persists settings for search facets

thostetler:
Adds a new `settings` slice to the store.  Although we may want to break it up again since the search facets required a good deal of handlers.  

Adds persistence of ordering, expanded, and open/closed state of all filters.

Adds a hide button on each filter and allows for showing them all again.

facet_hideable

### DIFF
--- a/src/pages/search/index.tsx
+++ b/src/pages/search/index.tsx
@@ -35,8 +35,8 @@ import { GetServerSideProps, NextPage } from 'next';
 import dynamic from 'next/dynamic';
 import Head from 'next/head';
 import { useRouter } from 'next/router';
-import { last, not, omit, path } from 'ramda';
-import { FormEventHandler, useEffect, useState } from 'react';
+import { last, omit, path } from 'ramda';
+import { FormEventHandler, useEffect } from 'react';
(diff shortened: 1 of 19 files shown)


src/pages/search/index.tsx
```diff
@@ -128,32 +128,31 @@ const SearchPage: NextPage = () => {
...
-          <title>{params.q} | NASA Science Explorer - Search Results</title>
-        </Head>
         {showFilters ? (
           <GridItem as="aside" aria-labelledby="search-facets">
             <Flex>
               <Heading as="h2" id="search-facets" fontSize="sm" flex="1">
                 Filters
               </Heading>
-              <Button variant="link" onClick={handleToggleFilters}>
+              <Button variant="link" onClick={() => handleToggleFilters()}>
```
shinyichen, line 155: Just curious why this was changed? Does `onClick={() => handleToggleFilters()}` not work here?

shinyichen (approved):
The new features are great!

src/pages/search/index.tsx
thostetler: not sure, but probably I had added some other call there and then removed... I'll fix!

PR #251, opened 2022-10-25
Print layout for results page

shinyichen:
Hide header, footer, search bar and facets in print layout

Screen Shot 2022-10-25 at 4 57 11 PM


thostetler (approved):
LGTM, although there is a conflict it says

PR #254, opened 2022-11-03
Scix 139 results graph

shinyichen:
## Component hierarchy
* `pages/search/results_graph` - the page for results graph
  * `components/visualizations/Containers/ResultsGraphPageContainer` - fetch and process the data for bubble plot
    * `components/visualizations/GraphPanes/BubblePlotPane` - containing BubblePlot, options and filters
      * `components/visualizations/Graphs/BubblePlot` - the actual graph

https://user-images.githubusercontent.com/636361/200414904-c383be78-5f92-4f6b-b6d8-f862dec33007.mov


thostetler (approved):
Nice work, this looks good and works pretty well

I was wondering though, if this needed to be fully implemented in custom d3? 
There weren't any plots that could be built on or customized to have this behavior?

My only concern being maintenance, since small bug can creep into these big graphs that might be quite difficult to track down. 

Also we should do a follow up PR to move the resulting `docs(xxx)` query into the FQ params.
I can help with that.

thostetler:
Makes sense!

PR #256, opened 2022-11-14
Update filter search in graphs to use fq

shinyichen:
https://user-images.githubusercontent.com/636361/201776981-32e9d9ee-32d4-4034-91a1-0ecd85d7afc9.mov


thostetler (approved):
This looks good, 

I definitely have to work with the types.  It's the inclusion of arbitrary keys to the request that is causing the problem.

PR #262, opened 2022-11-22
Scix 209 facet UI - implement drag & drop between visible and hidden areas

shinyichen:
* Visible list is sortable
* Hidden list is sorted alphabetically

https://user-images.githubusercontent.com/636361/203373651-a88ebe5d-23d4-487f-9d53-bbf7480128b0.mov

### DIFF
--- a/src/components/SearchFacet/SearchFacet.tsx
+++ b/src/components/SearchFacet/SearchFacet.tsx
@@ -23,9 +23,7 @@ import {
   MouseSensor,
   useDroppable,
   useSensor,
-  useSensors,
 } from '@dnd-kit/core';
-import { restrictToVerticalAxis } from '@dnd-kit/modifiers';
 import { arrayMove, SortableContext, useSortable, verticalListSortingStrategy } from '@dnd-kit/sortable';
 import { CSS } from '@dnd-kit/utilities';
 import { ExclamationCircleIcon, EyeIcon, EyeOffIcon } from '@heroicons/react/solid';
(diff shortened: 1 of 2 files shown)


shinyichen:
I made hidden list expand when item is dragged into, so you can see what's going on now.

> Look great so far. One small thing I would say is that when the hidden menu is closed, and you drag down the item just disappears. Maybe would be better to make it slide or fade away to make it visually clearer it's getting deleted. But let me know what you think

src/components/SearchFacet/SearchFacet.tsx
```diff
@@ -215,42 +252,163 @@ export const SearchFacets = (props: ISearchFacetsProps) => {
...
+      setShowHiddenFacets(true);
+    }
+    // item moved into hidden container when the list is empty
+    else if (over.id === 'hidden-container') {
+      if (hiddenFacets.length === 0) {
+        // move item to hidden
+        setFacetsList({
+          visible: visible.filter((id) => id !== active.id),
+          hidden: [active.id as SearchFacetID],
+        });
```
thostetler, line 276:
```suggestion
        setFacetsList((prev) => ({
          visible: prev.visible.filter((id) => id !== active.id),
          hidden: [active.id as SearchFacetID],
        }));
```

src/components/SearchFacet/SearchFacet.tsx
```diff
@@ -215,42 +252,163 @@ export const SearchFacets = (props: ISearchFacetsProps) => {
...
-          <List>{list}</List>
-        </SortableContext>
-      </DndContext>
-      {hidden.length > 0 && (
-        <Button variant="link" onClick={handleShowAllClick} type="button">
-          Show all ({hidden.length} hidden)
+        <DroppableContainer id="visible-container" itemsID={facetsList.visible} items={visibleItems} />
+
+        <Button variant="link" onClick={toggleShowHidden} type="button">
+          {showHiddenFacets ? 'Hide hidden filters' : 'Show hidden filters'} {`(${hiddenItems.length})`}
```
thostetler, line 371: I wonder if here it may be better to also have a chevron type toggle so that it's a clear indicator of the menu being open. 

When it's empty, you don't immediately know if it's open or not unless you actually read the text

thostetler (approved):
I think this looks pretty good to me.  
Couple of small suggestions.

For `setState` calls with an object, it is a little more flexible to pass a function and spread over it.  If we ever add more props there it's less to change later.

src/components/SearchFacet/SearchFacet.tsx
shinyichen: Added

PR #269, opened 2022-12-14
SCIX-208 fix bug where selecting group 1 shows outline around all groups

shinyichen:
# The problem
In paper network, sometimes selecting a group from the pie will results in highlighting of the whole pie. As seen in the screen capture.
Screen Shot 2022-11-10 at 4 10 56 PM
# The cause
The pie is created using a fake 'root' with id=0. When the actual data group also has a group of id=0, selecting group 0 will also select the root.
# The solution
Changed the fake root id to an unlikely number.


thostetler (approved):
👍

PR #268, opened 2022-12-13
Scix 225 histogram slider expand

shinyichen:
Depends on #267 

* Moved facets under search bar
* Allow histogram to expand

https://user-images.githubusercontent.com/636361/207434849-abd039ee-ed4c-4f2f-81b7-b746a963c8ed.mov

PR #271, opened 2022-12-29
Scix 231 slider component

shinyichen:
Depends on #268 
Refactor out histogram slider's slider into a component, and replace Chakra slider with it.


thostetler (approved):
LGTM

PR #272, opened 2022-12-30
SCIX 222 implement author affiliations

thostetler:
- Update component script template and add error helper utility
- Upgrade axios to 1.2.1
- Remove unused libs, upgrade MSW
- Minor updates
- linting fixes
- Add author-affiliation api entry
- Add Author Affiliations component
- Add author aff search size to default config
- Add mock handler for author affiliations
- Add functionality to author affiliations button in list actions
- Updates to hook useDownloadFile


thostetler:
@shinyichen there is a bug with modals that should be fixed when we upgrade Chakra in the other PR.  However, this PR isn't blocked by the upgrade

shinyichen:
@thostetler Still looking at the code, but just noticed a couple of issues I saw:
1. The first time I click on the author affiliation item in the dropdown doesn't take me there, I always have to click second time to work.
2. exported file content seems like html

The other thing is I am not sure what the max author does, doesn't match the number of authors in the list. But this is the same way in BBB.

thostetler:
@shinyichen I think it has to do with how many authors it takes from each work, but there may be a `max` value that the service retrieves that's less than the total number.  I'm not sure either though, have to look at the affilation_service.
And yeah regarding that bug, I'm able to reproduce -- I think it seems to only happen if I have no items selected (doing a query-based search)
I'll fix

thostetler:
Can you show me what exported content you're getting? the exports all look good to me

shinyichen:
This is the exported file. I believe I did search `author:"sagan, carl"` and used all the defaults.
[authoraffiliation_2023-01-05T10_38_10-05_00.csv](https://github.com/adsabs/nectar/files/10353822/authoraffiliation_2023-01-05T10_38_10-05_00.csv)

thostetler:
I'm unable to reproduce, hmm
make sure you did a `yarn`? 
I changed the `useDownloadFile` hook to use `file-saver`

thostetler:
I made a small change to make the hook keep previous data -- I think it makes changing params a little smoother

shinyichen (approved):
The two issues I mentioned are gone now.

PR #273, opened 2022-12-30
Scix 232 histogram slider bug

shinyichen:
Depends on #271 
### Reproduce
search star then star galaxy then star galaxy chemical tagging and then star galaxy
Now the year range is applied to the year range of star galaxy chemical tagging. 
### Solution
Use slider's `onSlideEnd` (which only update when handles dragged by use) instead of `onChange` (updates both when user drags and component input prop is updated). 

### Reproduce
If no results or results only contains one year, histogram is broken.
### Solution
* If no results, hide facets. 
* If results only has 1 year, slider does not allow, so hide slider and only show the label for that year.
Screen Shot 2022-12-30 at 10 11 12 AM


thostetler (approved):
:+1:

PR #275, opened 2023-01-04
Replace range query with fq

shinyichen:
In explore -> overview, when year filter is applied, use fq instead of range query


thostetler:
@shinyichen, So I threw another commit on here.  It's annoying that we have to keep casting the types back to `Query` when working with this query-utils module.
I updated `IADSApiSearchParams` type to allow arbitrary keys, so now it should just work with the query-utils.  I think I got all the areas that are breaking, but let me know if you see more.

thostetler (approved):
The changes you made look good, please check out the commit I added which includes some typings updates

shinyichen:
> The changes you made look good, please check out the commit I added which includes some typings updates

Awesome, thank you.

PR #278, opened 2023-01-09
fix quick fields bug

thostetler:
- use a flag for updating cursor in search bar


shinyichen (commented):
The only issue I saw was when clicked on a search example, the cursor also goes back. Otherwise no problem.

thostetler:
I'm unable to reproduce actually, could you try again for me

PR #281, opened 2023-01-19
SCIX 227 create new dialogs for facets based on jc mockups

thostetler:
image

image

facet_modal


shinyichen:
Pagination issue:
* Search `star` and select `more` on authors
* Go to a different page, say '6'
* Change sort to alphabetical, and select `A`
* I think page is not reset here, because results are not showing

shinyichen:
An interesting thing when selecting a root item that has been expanded. The children are also selected.
https://user-images.githubusercontent.com/636361/215505740-c86a5547-0fe0-479f-8715-00941d13d224.mov

shinyichen:
When viewing by a particular alphabet and switch back to count view, items in the list are not reset to all items

https://user-images.githubusercontent.com/636361/215570968-44125e91-8b03-4f0e-bfb6-60038e204a15.mov

thostetler:
> An interesting thing when selecting a root item that has been expanded. The children are also selected.
> 
>  215505740-c86a5547-0fe0-479f-8715-00941d13d224.mov

Regarding this, I'm not sure what would be the best to do here, since this is similar behavior to how Bumblebee is doing it.
I'm thinking probably we want to keep the root item ONLY if it's selected explicitly, but how to handle if only a subset of the children are selected?

This was a problem before this modal

What do you think would be the best way to handle this?

shinyichen:
If selecting just the root or when all children are selected, I would expect only the root is shown in the list. And when filter is applied, you only see one filter (right now the result is all filters combined)

I would expect when a subset of children is selected, then the results is all the selected one combined.

Here I tried using BBB, it seems to behave like what I described, but it assumes you are selecting all when you have selected all the loaded children. 


https://user-images.githubusercontent.com/636361/215599575-0da5cdd1-b195-4258-a3ca-6d4eb3895421.mov



> > An interesting thing when selecting a root item that has been expanded. The children are also selected.
> > 215505740-c86a5547-0fe0-479f-8715-00941d13d224.mov
> 
> Regarding this, I'm not sure what would be the best to do here, since this is similar behavior to how Bumblebee is doing it. I'm thinking probably we want to keep the root item ONLY if it's selected explicitly, but how to handle if only a subset of the children are selected?
> 
> This was a problem before this modal
> 
> What do you think would be the best way to handle this?

shinyichen (commented):
No other comments besides the ones already mentioned.

thostetler:
> If selecting just the root or when all children are selected, I would expect only the root is shown in the list. And when filter is applied, you only see one filter (right now the result is all filters combined)

@shinyichen Thanks, I agree with you -- 
The only other case I would wonder about is if:
1. User selects and expands an author like `Wang, J` which has many children.
2. We only see a subset, those are automatically selected (ex. 10)

The user doesn't technically have ALL the children selected -- only the first 10 from the search.  In Bumblebee, the behavior is to _assume_ the user means they want all (so it selects the root, and doesn't care about the children)

I think probably this is fine, since I wouldn't expect a user means to select the subset and not the rest.

shinyichen:
A few issues while playing with the latest update:
* Select a second level filter, and go back to parent, parent checkbox is no longer 'intermediate'  (this was working before)
* Select parents, children are not automatically marked checked (not sure if this was intentional)
* Alphabetical sort doesn't seem to do anything except in Authors
* Some facets have multiple list of selected keywords
Screen Shot 2023-02-27 at 10 52 15 AM

PR #286, opened 2023-02-07
Prevent unnecessary repaint of histogram.

shinyichen:
While loading the results page and there is no query data available, the histogram should still be allowed to be shown. This way histogram will not disappear when new filters are applied.


thostetler (approved):
LGTM

PR #288, opened 2023-02-28
Scix 212 settings 2

shinyichen:
Second round of work on settings:
* Token setting page
* Rework on the export settings UI
* Refactor into smaller UI components


shinyichen:
@marblestation This PR contains the new UI for the export settings. Feel free to test it and give me any feedbacks.

src/components/Settings/Export/Tabs/BibtexTabPanel.tsx
```diff
@@ -0,0 +1,246 @@
...
+  // If bibtex settings and bibtex Abs settings are the same
+  // we can use basic view
+  const canUseBasicMode = useMemo(
+    () =>
+      bibtexKeyFormat === bibtexABSKeyFormat &&
+      bibtexAuthorCutoff === bibtexABSAuthorCutoff &&
+      bibtexMaxAuthors === bibtexABSMaxAuthors &&
+      bibtexAuthorCutoff === bibtexMaxAuthors, // basic settings cutoff and max authors are the same
+    [],
+  );
```
thostetler, line 43: probably unnecessary memoization, since these are pretty light operations/logic

src/components/Settings/Export/Tabs/BibtexTabPanel.tsx
```diff
@@ -0,0 +1,246 @@
...
+  }, [
+    bibtexJournalFormat,
+    bibtexKeyFormat,
+    bibtexAuthorCutoff,
+    bibtexMaxAuthors,
+    bibtexABSKeyFormat,
+    bibtexABSAuthorCutoff,
+    bibtexABSMaxAuthors,
+    selectedTab,
+  ]);
```
thostetler, line 68: Same here, I'd remove the memo since this is all basic logic, you can roll it into a helper file or something if you want to clean up the component a bit, but this should be fine to leave here if you'd like

src/pages/user/settings/token.tsx
```diff
@@ -1,14 +1,33 @@
...
+  useEffect(() => {
+    if (data) {
+      setToken(data.access_token);
+    }
+  }, [data]);
 
   const handleGenerateToken = () => {
-    console.log('generate new token');
+    (async () => {
+      const res = await generateNewApiToken();
```
thostetler, line 24: This page will require more than this to keep in sync -- I can work on writing up a handler for this in the useSession hook

thostetler (changes requested):
I think the changes look good -- I added a commit that improves the merging of the store state, which should help with some random hydration issues I was getting.
 
One thing I noticed is that you should try to test if the `useMemo` is needed, like if it's slow without it -- since it might be a premature optimization.

I'll work on adding a generate token handler in the `useSession` hook for the token page.  We'll want to roll the pages I worked on for changing email/password into this as well.  This can all happen in future PRs

src/pages/user/settings/token.tsx
shinyichen: Ya, let me know what I need to do here.

thostetler: Hi, yeah if you are ready go ahead and merge -- we can work on that as a separate PR. I have a branch off here to start (https://github.com/adsabs/nectar/compare/master...thostetler:nectar:token_settings)
But presumeably we'll need to rethink about it since I've since changed the way we handle the sessions slightly

thostetler (approved):
These changes look good to me

PR #293, opened 2023-04-05
Minor UI issues

shinyichen:
### 1 Spacing while loading
Make the loading skeleton size consistent with content, so that there is less (or no) shifting of content switching from content to loading

https://user-images.githubusercontent.com/636361/230126132-ca53bb66-9da1-426b-988f-02501e1792fe.mov

### 2 Update the filter list to horizontal
Screen Shot 2023-04-05 at 11 18 03 AM

### 3 Histogram: if only one year
Make it look a little more like a histogram by changing the bar width. Not appending more years because this might happen when using `year:` in the query, and the histogram implies other years were empty, which would be wrong.

Screen Shot 2023-04-05 at 11 19 29 AM

### 4 All search term dropdown
Make it more consistent with the rest of the page
Screen Shot 2023-04-05 at 11 22 09 AM

### 5 For printing, remove checkboxes and controls
Screen Shot 2023-04-05 at 11 23 28 AM

### DIFF
--- a/src/components/NumFound/NumFound.tsx
+++ b/src/components/NumFound/NumFound.tsx
@@ -1,6 +1,6 @@
 import { useGetSearchStats } from '@api';
 import { Text } from '@chakra-ui/layout';
-import { SkeletonText } from '@chakra-ui/react';
+import { Box, SkeletonText } from '@chakra-ui/react';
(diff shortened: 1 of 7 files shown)


thostetler (approved):
Changes LGTM

PR #298, opened 2023-04-27
Use user setting in citation export

shinyichen:
# SCIX-250 Allow user to choose specific custom format as a default format in export setting
Disable user sorting custom formats, and move a custom format to first place only if that is chosen as the default format.

Screen Shot 2023-04-27 at 11 57 24 AM

# SCIX-251 Support custom format in citation export

Screen Shot 2023-04-27 at 11 58 27 AM

Screen Shot 2023-04-27 at 11 59 45 AM

# SCIX-248 Use user setting in citation export if logged in
Also use bibtex and bibtex abs settings from user setting.
Support basic (maxauthor only) and advance (maxauthor and cutoff) mode for bibtex and bibtex abs

https://user-images.githubusercontent.com/636361/234919835-780171ad-bfed-4820-b012-539724a362b5.mov


src/components/CitationExporter/components/CustomFormatSelect.tsx
```diff
@@ -1,10 +1,93 @@
...
+  // custom formats to options
+  const customFormatOptions = customFormats
+    .map((f) => ({
+      id: f.id,
+      label: f.name,
+      value: f.id,
+      code: f.code,
+    }))
+    .sort((a, b) => (a.label < b.label ? -1 : 1));
+  customFormatOptions.splice(0, 0, { id: 'new', label: 'Enter New Format', value: 'new', code: '%1H:%Y:%q' });
```
thostetler, line 24: This "enter new format" makes me think that it's going to save it somewhere for me.  maybe "Enter custom format" or something like that.  Unless the plan is to allow users to save a format from here

thostetler (changes requested):
Looks pretty good, had a couple questions/suggestions

src/components/CitationExporter/components/CustomFormatSelect.tsx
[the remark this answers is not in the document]
shinyichen: Changed to `unshift`.

shinyichen: Updated.

shinyichen:
@thostetler I've addressed your feedback in the latest commit.

thostetler:
Changes look good, please merge when ready

thostetler (approved):

PR #310, opened 2023-06-28
Updates to ORCiD

thostetler:
Fix up orcid profile page
Adds some tests
Improves caching invalidation/update strategy when adding/deleting works
Updates navbar to make actual switch
Improves error handling, using suspense for both the table and the settings
Improves responsiveness


shinyichen:
Is this correct? There are some articles in SciX but show up in the 'Not in SciX' list
Screen Shot 2023-07-10 at 10 00 27 AM

shinyichen:
I get this error message the first time I select from the dropdown. 
Screen Shot 2023-07-10 at 10 23 16 AM

shinyichen:
Just playing around with the PR and noticed these things:
* Logout from menu doesn't seem to work unless in the 'My orcid page' 
* Bulk claim: I got an error `There was an error when updating the record. Please try again. If the error persists, please contact NASA Astrophysics Data System for assistance.` but all the items turned to 'claimed' 
* Bulk delete (10 items): Two 'successful' toasts shown.
* After bulk claim, change page does not work correctly, the items stay the same

thostetler:
> I get this error message the first time I select from the dropdown. Screen Shot 2023-07-10 at 10 23 16 AM

I looked into this and can't quite figure out how to even catch this one, wrapping the mathjax component with an error boundary doesn't work.  I believe this is because it happens inside a promise. 

I think what's happening is related to the pagination, the table pre-renders all cells, then only shows the first 10 or whatever, and then the promises come back and attempt to update the dom - but the element doesn't exist.  So I think for now we'll ignore the error, until a solution is found -- I don't think there are any performance issues or anything that is buggy because of this.

thostetler:
@shinyichen thanks!
I've made fixes for all the mentioned items.  I think I'll merge this in and we can then refactor as necessary since this one is getting big.

PR #311, opened 2023-07-03
Feedback UI SCIX-269

shinyichen:
* Added page for feedback forms
  * general
  * associated articles
  * missing references
  * missing record
* Form validation
* Components that make up the pages

Screen Shot 2023-07-03 at 2 06 17 PM


src/pages/feedback/general.tsx
```diff
@@ -0,0 +1,113 @@
...
+      <Formik initialValues={initialFormValues} onSubmit={handleSubmitForm}>
+        {(props) => (
+          <Form>
+            <Flex direction="column" gap={4}>
+              <HStack gap={2}>
+                <Field name="name">
+                  {({ field }: FieldProps) => (
+                    <FormControl isRequired>
+                      <FormLabel>Name</FormLabel>
+                      <Input {...field} />
```
thostetler, line 76: should focus upon entering page

src/pages/feedback/general.tsx
```diff
@@ -0,0 +1,113 @@
...
+  HStack,
+  useDisclosure,
+  AlertStatus,
+} from '@chakra-ui/react';
+import { FeedbackLayout, FeedbackAlert } from '@components';
+import { useStore } from '@store';
+import { Field, FieldProps, Form, Formik, FormikHelpers } from 'formik';
+import { NextPage } from 'next';
+import { useState } from 'react';
+
```
thostetler, line 18: All these pages should export that default getServerSideProps
```typescript
export { injectSessionGSSP as getServerSideProps } from '@ssrUtils';
```
unless they implement it.

src/pages/feedback/general.tsx
```diff
@@ -0,0 +1,113 @@
...
+type FormValues = {
+  name: string;
+  email: string;
+  feedback: string;
+};
+
+const General: NextPage = () => {
+  const username = useStore((state) => state.getUsername());
```
thostetler, line 26: I'd say rather than this, just grab the username from the user object. 

```typescript
const username = useStore(state => state.user.username);
```

src/components/FeedbackForms/MissingRecord/RecordPanel.tsx
```diff
@@ -0,0 +1,232 @@
...
+            <FormLabel>Collection</FormLabel>
+            <CheckboxGroup>
+              <Field name="collection">
+                {({ field }: FieldProps) => (
+                  <Stack direction="row">
+                    {collections.map((c) => (
+                      <Checkbox key={`collection-${c.value}`} {...field} value={c.value}>
+                        {c.label}
+                      </Checkbox>
+                    ))}
```
thostetler, line 112: Could probably not bother with the config and loop pattern here, since it's unlikely to change frequently, we can just list out the three checkboxes to make it simpler.

thostetler (approved):
This looks pretty good so far.  I noticed a few things.

1. Make sure that we properly focus when entering a page, also after adding something new/editing focus should update appropriately.

2. Need to add the default GSSP export on all pages, I mention it in another comment, so that we keep server state in sync.

3. The preview is a little barebones at the moment, I think though this approach is probably fine.  The original-original form from classic attempted to diff the changes, which is also what I try to do, but it's not fool-proof.  Maybe this is preferred.

4. The date picker is possibly unnecessary since it's often only a month and year.  I'd say we could just do some light validation on the field and give helper text.

I think maybe the checkboxes could be justified or stretched out a bit.
image

thostetler:
One more thing I noticed is the form is sometimes a little laggy.  Certain fields rerender the whole form on change, we may want to try to isolate some of that.  But we'll see, maybe the performance is okay.

https://github.com/adsabs/nectar/assets/6970899/74b52ff1-29e1-44b5-8de5-363b66cda718

src/components/FeedbackForms/MissingRecord/RecordPanel.tsx
shinyichen: I used the list since for the expansion, I don't know what the collections look like. But for the next phase, I changed the type of the `database` field in the search response, instead of keeping a list here.

shinyichen:
I'm leaving items related to the missing record page for the next PR because it has been changed quite a bit. 
* Improve the preview with diff view
* Remove date picker
* Refactor and make form faster

PR #314, opened 2023-07-28
get storybook 7 working

thostetler:
- Moves quickfields to their own module, uses indirect updates to change query
- Adds flags to store for updating query, they will allow a string to be appended or the query cleared.  This allows us to know that something was changed from outside the component, since we're always updating the query.
- Small updates to storybook, and clean up
- get storybook 7 working
- updating chakra imports
- Fixes to searchbar


thostetler:
https://github.com/adsabs/nectar/pull/315 should be done first, this PR builds on it.

shinyichen (approved):

PR #329, opened 2023-09-26
add planetary search demo

thostetler:
https://github.com/adsabs/nectar/assets/6970899/ffd2e932-0abb-4b0b-9978-f2afaa3d2c11

Creates a new "experiments" menu for this type of thing.  Don't know if we'll have more, but it's now setup :smile:


thostetler:
@shinyichen This is a one-off demonstration of this planetary search that @golnazads put together.  I'm just shoving it into nectar as an "experiment."  Let me know if you see any issues, thanks!

PR #330, opened 2023-10-04
Fix highlight result ordering

shinyichen:
In search results, highlights were mapped to the wrong articles because `highlightings` is an object, and has no ordering. This fix maps highlights to array in the same order as the search results first.


thostetler (approved):
LGTM

PR #336, opened 2023-10-25
My Libraries

shinyichen:
All the existing libraries implementations
* Settings: update metadata, change ownership, add collaborators, delete library
* Adding/Deleting papers to library from search results and abstract page
* Public library view
https://github.com/adsabs/nectar/assets/636361/37487083-b9d8-4325-a1e1-1475132efc3b


thostetler:
image
I think we should hide the pagination controls and operations when no libraries are found.

thostetler:
Small nits:
image
should lowercase the 'a'

---

image
I think we could have a little more vertical margin, feels a little closed in when you have a low amount of libs

---

https://github.com/adsabs/nectar/assets/6970899/5b2d92e9-539e-4702-a4f2-614027cba5a3
The back button on a paper details page doesn't go back to main libraries page as I would expect

---

image
Seems like collaborators are slightly off center

---

image
I think I would prefer the modal to be bigger so the library list isn't quite so tight

---

https://github.com/adsabs/nectar/assets/6970899/17aaa5c3-a587-486c-9089-aa62d139cafc
Seeing a bug where after adding a new library it's not showing up in the list

---

image
Should use humanized date strings, similar to how we have it in the ORCiD works table:
https://github.com/adsabs/nectar/blob/2281f597f3eab27d5fa74330694f56095ad9399f/src/components/Orcid/WorksTable.tsx#L304

---

image
On library details page, when request is blocked (error state) the page remains in a loading state

---

image
on the main libraries table, if `biblib` requests are blocked, the page throws an error

---

image
I'm not clear on the purpose of the dropdown at the top.

src/components/__tests__/LibraryListTable.test.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { test } from 'vitest';
+import { composeStories } from '@storybook/testing-react';
+import * as stories from '../__stories__/LibraryListTable.stories';
+import { ChakraProvider } from '@chakra-ui/react';
+
+const { Default: LibraryListTable } = composeStories(stories);
+
+test('renders without crashing', () => {
+  render(
+    <ChakraProvider>
```
thostetler, line 11: this provider could be added to test-utils https://github.com/adsabs/nectar/blob/56d14cdfed266797b90d3e737fa5474cdeb2c506/src/test-utils.tsx#L63

src/components/Libraries/LibraryListTable.tsx
```diff
@@ -0,0 +1,333 @@
...
+  hideCols?: Column[];
+  showDescription?: boolean;
+  onChangeSort: (sort: ILibraryListTableSort) => void;
+  onChangePageIndex: (index: number) => void;
+  onChangePageSize: (size: NumPerPageType) => void;
+  onLibrarySelect: (id: LibraryIdentifier) => void;
+  onUpdate?: () => void;
+}
+
+export const LibraryListTable = (props: ILibraryListTableProps) => {
```
thostetler, line 108: Instead of rolling your own table here, you can use the react-table lib I added for the ORCiD works table.  Since it handles sorting/ordering/pagination/etc

src/components/Libraries/LibrarySelector.tsx
```diff
@@ -0,0 +1,126 @@
...
+          <Stack dir="column">
+            {selected.map((l) => (
+              <InputGroup key={l.name}>
+                <Input value={l.name} isReadOnly />
+                <InputRightElement>
+                  <IconButton
+                    icon={<CloseIcon />}
+                    aria-label="Remove"
+                    colorScheme="gray"
+                    variant="ghosted"
```
thostetler, line 94: Isn't it `ghost`?

src/components/Libraries/LibrarySettingsPane.tsx
```diff
@@ -0,0 +1,281 @@
...
+    );
+  };
+
+  return (
+    <>
+      <Box>
+        <Heading variant="pageTitle" as="h1" my={4}>
+          <SimpleLink href={`/user/libraries/${id}`} display="inline">
+            <ChevronLeftIcon mr={2} />
+          </SimpleLink>
```
thostetler, line 162: We should have a component for these back links.  This should work like a stack that we 'push' as we dig in, and 'pop' as we go back up

src/components/Libraries/OperationModal.tsx
```diff
@@ -0,0 +1,272 @@
...
+  const action = useWatch<FormValues, 'action'>({ name: 'action', control });
+
+  const handleAddLib = (id: LibraryIdentifier) => {
+    append({ value: id });
+  };
+  const handleRemoveLib = (id: LibraryIdentifier) => {
+    remove(libs.findIndex((l) => l.value === id));
+  };
+
+  const handleOperate = () => {
```
thostetler, line 145: I believe this method already gets passed the form values

src/pages/user/libraries/[[...id]].tsx
```diff
@@ -1,18 +1,118 @@
...
+    },
+    { enabled: !!id },
+  );
+
   return (
-    <div>
+    <>
       <Head>
-        <title>NASA Science Explorer - Libraries</title>
+        <title>NASA Science Explorer - Libraries - {!!id ? library?.metadata.name ?? '' : ''}</title>
```
Comment (thostetler) [inline src/pages/user/libraries/[[...id]].tsx:37]: For dynamic titles we have to put the whole thing as a template string, otherwise it generates a `<!-- -->` comment string in the title

src/pages/user/libraries/[[...id]].tsx
```diff
@@ -1,18 +1,118 @@
...
+      )}
+    </>
   );
 };
 
 export default LibrariesHome;
-export { injectSessionGSSP as getServerSideProps } from '@ssr-utils';
+
+export const getServerSideProps: GetServerSideProps = composeNextGSSP(async (ctx) => {
+  if (!ctx.req.session.isAuthenticated) {
```
Comment (thostetler) [inline src/pages/user/libraries/[[...id]].tsx:77]: This is unnecessary, this check is done automatically in the middleware

src/utils.ts
```diff
@@ -496,3 +496,10 @@ export const coalesceAuthorsFromDoc = (doc: IDocsEntity, includeAff?: boolean) =
 export const pluralize = (str: string, count: number) => {
   return count === 1 ? str : `${str}s`;
 };
+
+export const isValidEmail = (email: string) =>
```
thostetler, line 500: What's the benefit of using this regex vs. using the `email` check from `zod`?

thostetler (approved):
Nice work, everything works well and feels natural coming from BBB.  I have a few things I noticed, but no show stoppers.
LGTM

src/components/Libraries/LibraryListTable.tsx
shinyichen: I was looking at that before, but my table rows are fetched on the fly, and the sorting is a params in the query.

shinyichen:
> image I think we should hide the pagination controls and operations when no libraries are found.

Resolved

shinyichen:
Resolved.

The drop down is a placeholder where we will have 'my libraries', 'shared with me' and 'following' in the future.

> image should lowercase the 'a'
> 
> image I think we could have a little more vertical margin, feels a little closed in when you have a low amount of libs
> 
> image We could probably keep the modal open and let the user know that the name they chose was already taken. I'm thinking of a situation where I entered the wrong name, but wrote out a long description -- currently I'd lose what I wrote...
> 
> https://github.com/adsabs/nectar/assets/6970899/5b2d92e9-539e-4702-a4f2-614027cba5a3 The back button on a paper details page doesn't go back to main libraries page as I would expect
> 
> image Seems like collaborators are slightly off center
> 
> image I think I would prefer the modal to be bigger so the library list isn't quite so tight
> 
> https://github.com/adsabs/nectar/assets/6970899/17aaa5c3-a587-486c-9089-aa62d139cafc Seeing a bug where after adding a new library it's not showing up in the list
> 
> image Should use humanized date strings, similar to how we have it in the ORCiD works table:
> 
> https://github.com/adsabs/nectar/blob/2281f597f3eab27d5fa74330694f56095ad9399f/src/components/Orcid/WorksTable.tsx#L304
> 
> image On library details page, when request is blocked (error state) the page remains in a loading state
> 
> image on the main libraries table, if `biblib` requests are blocked, the page throws an error
> 
> image I'm not clear on the purpose of the dropdown at the top.
> 
> https://github.com/adsabs/nectar/assets/6970899/130067b3-f4bb-4b62-8829-46c2e637265c Deleting a library from settings does not update the main library list.
> 
> Other small nits:
> 
> * In the modals that are in use, we should place focus into whatever primary input/button



> image should lowercase the 'a'
> 
> image I think we could have a little more vertical margin, feels a little closed in when you have a low amount of libs
> 
> image We could probably keep the modal open and let the user know that the name they chose was already taken. I'm thinking of a situation where I entered the wrong name, but wrote out a long description -- currently I'd lose what I wrote...
> 
> https://github.com/adsabs/nectar/assets/6970899/5b2d92e9-539e-4702-a4f2-614027cba5a3 The back button on a paper details page doesn't go back to main libraries page as I would expect
> 
> image Seems like collaborators are slightly off center
> 
> image I think I would prefer the modal to be bigger so the library list isn't quite so tight
> 
> https://github.com/adsabs/nectar/assets/6970899/17aaa5c3-a587-486c-9089-aa62d139cafc Seeing a bug where after adding a new library it's not showing up in the list
> 
> image Should use humanized date strings, similar to how we have it in the ORCiD works table:
> 
> https://github.com/adsabs/nectar/blob/2281f597f3eab27d5fa74330694f56095ad9399f/src/components/Orcid/WorksTable.tsx#L304
> 
> image On library details page, when request is blocked (error state) the page remains in a loading state
> 
> image on the main libraries table, if `biblib` requests are blocked, the page throws an error
> 
> image I'm not clear on the purpose of the dropdown at the top.
> 
> https://github.com/adsabs/nectar/assets/6970899/130067b3-f4bb-4b62-8829-46c2e637265c Deleting a library from settings does not update the main library list.
> 
> Other small nits:
> 
> * In the modals that are in use, we should place focus into whatever primary input/button

src/components/Libraries/LibrarySettingsPane.tsx
shinyichen: What's the advantage for the component? I can add a task for this.

thostetler: Yeah we could think about it anyway, we should probably use hash urls or something like that as well for the sub-views so that we could keep track of it/shareable/etc.

PR #337, opened 2023-10-27
General fixes 1

thostetler:
- Add a HideOnPrint utility component
- Fix bug with create-component script
- Switch over any lingering Tailwindcss class usages
- Remove tailwindcss, postcss, and autoprefixer
- Update all global tailwind classes to css equivalents
- Remove TypeaheadMenu.tsx, not used
- Fix focus ring around compound-components
- Add status:rejected to works table filter
- Improve visibility of clear button on facet selection list
- get husky lint-staged working again
- Remove momentjs, slight fix to parsePubDate utility


src/components/ResultList/Item/Item.tsx
```diff
@@ -177,13 +179,13 @@ const Highlights = ({
   }, [highlights]);
 
   return (
-    <Box as="section" aria-label="highlights" className="search-snippets" my={2} data-testid="highlights-section">
+    <Box as="section" aria-label="highlights" fontStyle="italics" my={2} data-testid="highlights-section">
```
shinyichen, line 182: The highlighting color is gone, it's hard to tell where the highlights are

thostetler: Good catch, fixed! :+1:

PR #341, opened 2023-10-30
SCIX-290 Add related materials in abstract

shinyichen:
Add "Related Materials" for abstract
Screen Shot 2023-10-30 at 11 04 39 AM


thostetler (approved):
LGTM

PR #342, opened 2023-10-31
Add extra info to README.md

JCRPaquin:
Added some information I needed to get started with nectar development.

cc @thostetler @shinyichen

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@
 
 #### Dependencies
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,7 +14,18 @@ pnpm install
...
+Then you can start the development server with mock API endpoints:
+
+```bash
+pnpm dev:mocks
+```
+
+Or, run the development server against live APIs:
 
 ```bash
 pnpm dev
```
JCRPaquin, line 31: It looks like this requires an API proxy? Is there a command to run before this to ensure the dev frontend can access the backend?

PR #343, opened 2023-10-31
Sorting related items

shinyichen:
* **Scix 291**: The issue was originally just making second-order searches sort by score, but ended up with changing the default sort to score. Date is secondary (tie breaker).
* Reorder sort list with Score and Date on top
* Make show highlight button a switch so it is less prominent
* **Scix309**: Remove 'classic factor' and 'bibcode' from sort list

Screen Shot 2023-10-31 at 12 17 25 PM


thostetler:
My personal opinion would be to stick a toggle like highlights into a cog menu off to the right side, we could then add more later.  I don't think it's necessary to have it so prominent.  But we should get some opinions about it maybe

thostetler (approved):
Changes look good, I'm not a huge fan of this highlights switch, but it can stay like this for now, especially if the team is okay with it.

shinyichen:
@thostetler I added a menu, does it look better?

https://github.com/adsabs/nectar/assets/636361/31e3955a-7e3f-4c29-ad82-0557db013666

thostetler:
The menu looks good, although it seems a little transparent.  I think it's a better solution than the switch alone though. Thanks for changing :+1:

PR #345, opened 2023-10-31
Scix 314 Fix bugs related to visualization page

shinyichen:
* Fix tabs not working
* Fix undefined error
* Fix tab that is auto switching back


thostetler (approved):
LGTM

PR #348, opened 2023-11-06
fix author links

shinyichen:
Fix links not working
* author/orcid links in abstract
* author/orcid links in authors list
* "view as search results" link
* keywords links

### DIFF
--- a/src/pages/abs/[id]/abstract.tsx
+++ b/src/pages/abs/[id]/abstract.tsx
@@ -110,7 +110,7 @@ const AbstractPage: NextPage<IAbstractPageProps> = (props: IAbstractPageProps) =
                         params={createQuery('orcid', orcid)}
                         aria-label={`author "${author}", search by orKid`}
                       >
-                        <OrcidActiveIcon fontSize={'large'} />
+                        <OrcidActiveIcon fontSize={'large'} mx={1} />
                       </SearchQueryLink>
                     )}
                     <>{index === MAX - 1 || index === doc.author_count - 1 ? '' : ';'}</>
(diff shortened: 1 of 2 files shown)


src/pages/abs/[id]/abstract.tsx
```diff
@@ -110,7 +110,7 @@ const AbstractPage: NextPage<IAbstractPageProps> = (props: IAbstractPageProps) =
                         params={createQuery('orcid', orcid)}
                         aria-label={`author "${author}", search by orKid`}
                       >
-                        <OrcidActiveIcon fontSize={'large'} />
+                        <OrcidActiveIcon fontSize={'large'} mx={1} />
```
Comment (thostetler) [inline src/pages/abs/[id]/abstract.tsx:113]: I haven't tested this, but do these now show up in the tabindex? 
I wasn't able to tab into these icon-links.  We may want to change to `<IconButton>` and restyle.  But if not we'll need to add some extra attributes to make more accessible

thostetler (changes requested):

src/pages/abs/[id]/abstract.tsx
Comment (shinyichen) [reply src/pages/abs/[id]/abstract.tsx:113]: @thostetler Which browser? I see them as `<a>` tag and in Chrome I was able to tab through them. Firefox on Mac requires turning it on in the settings and Safari also has tabbing through links turned off by default.

thostetler (approved):

src/pages/abs/[id]/abstract.tsx
Comment (thostetler) [reply src/pages/abs/[id]/abstract.tsx:113]: Nevermind after testing these look good :+1:

PR #349, opened 2023-11-06
don't use the default user name when not logged in

shinyichen:
* Don't use the 'anonymous@ads' in feedback form
* Remove 'legacy' and 'career' from 'about' menu


thostetler (approved):
LGTM

PR #350, opened 2023-11-06
SCIX-240 disable click on disabled nav menu

shinyichen:
Fix abstract menu item clickable when it is disabled.


thostetler (approved):
LGTM

PR #351, opened 2023-11-07
Check for undefined doc when fetching

shinyichen:
Abstract pages were having 'undefined' errors if loading pages directly, while data is being fetched. This PR make sure those are checked.


thostetler (approved):
LGTM

PR #366, opened 2023-11-15
correct library modified date

shinyichen:


src/components/Libraries/LibraryListTable.tsx
```diff
@@ -338,8 +338,8 @@ export const LibraryListTable = (props: ILibraryListTableProps) => {
 };
 
 const LastModified = ({ date }: { date: string }) => {
-  const dateStr = new Date(date);
-  const formatted = intlFormatDistance(new Date(date), new Date());
+  const dateStr = new Date(`${date.match(/[+]([0-9]){2}:([0-9]){2}$/) ? date : `${date}+00:00`}`); // date string here is missing the timezone, add it or the time is wrong
```
thostetler, line 341:
```suggestion
  // date string here is missing the timezone, add it or the time is wrong
  const dateStr = new Date(`${date.match(/\+\d{2}:\d{2}$/) ? date : `${date}+00:00`}`); 
```

thostetler: slightly more simplified regex

thostetler (approved):
Small suggestion -- looks good otherwise :+1:

PR #382, opened 2023-12-01
Fix Details Page SSR Issue

thostetler:
Noticed that the abstract page, and subpages were not actually being SSR'd.  This was due to an addition to the HOC that caused dehydrated states to get overridden by previous ones (my fault).   This fixes the issue, and refactors all those GSSP calls into the HOC.

- Add some padding around abstract
- Properly merge dehydrated states when composing GSSP's
- Make graphics and metrics errors return instead of throwing
- Make resolver queries not consider 404 status as error
- Small cleanup/refactoring
- Consolidate all server-side calls


thostetler:
I'll have a followup with this one to improve the error handling on these pages.  Since I removed the error prop from each page.

shinyichen (approved):

PR #386, opened 2023-12-04
Scix 346 and Scix 334 paper network links bug

shinyichen:
A couple of fixes on paper network
* correct the group colors
* Bug caused by the fake node's `stable_id` conflicted with real node's `stable_id`


thostetler (approved):
LGTM

PR #392, opened 2023-12-08
Export citation related bugs

shinyichen:
Sometimes get user setting returns empty object, or some values are not present (never been set), they need to be set to default values.

### DIFF
--- a/src/components/Settings/Export/Tabs/GeneralTabPanel.tsx
+++ b/src/components/Settings/Export/Tabs/GeneralTabPanel.tsx
@@ -1,14 +1,13 @@
-import { CustomFormat, ExportApiFormatKey, IDocsEntity, useGetExportCitation } from '@api';
+import { CustomFormat, ExportApiFormatKey, IDocsEntity, useGetExportCitation, useGetUserSettings } from '@api';
 import { Stack } from '@chakra-ui/react';
 import { SampleTextArea } from '@components';
 import { ExportFormat, exportFormats } from '@components/CitationExporter';
-import { JournalFormatMap } from '@components/Settings/model';
+import { DEFAULT_USER_DATA, JournalFormatMap } from '@components/Settings/model';
 import { UserDataSetterEvent } from '@pages/user/settings/export';
 import { values } from 'ramda';
 import { Dispatch, useMemo } from 'react';
 import { ExportFormatSelect } from '../ExportFormatSelect';
 import { CustomFormatSelect } from '../CustomFormatSelect';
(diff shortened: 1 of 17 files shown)


src/components/Settings/Export/Tabs/GeneralTabPanel.tsx
```diff
@@ -1,65 +1,78 @@
...
-  const { settings: userSettings } = useSettings();
+  const { data } = useGetUserSettings({ initialData: DEFAULT_USER_DATA });
+
+  // fill any missing user data with default
+  const userSettings = useMemo(
+    () => (data ? { ...DEFAULT_USER_DATA, ...Object.entries(data).filter((s) => !!s) } : DEFAULT_USER_DATA),
+    [data],
+  );
+
+  const exportFormatOptions = values(exportFormats);
```
thostetler, line 27: This should already be happening inside of `useSettings`.  In there I'd update the `initialData` prop

src/pages/abs/[id]/exportcitation/[format].tsx
```diff
@@ -1,27 +1,38 @@
...
+    enabled: isAuthenticated,
+  });
+
+  // fill any missing user data with default
+  const settings = useMemo(
+    () =>
+      settingsData ? { ...DEFAULT_USER_DATA, ...Object.entries(settingsData).filter((s) => !!s) } : DEFAULT_USER_DATA,
+    [settingsData],
   );
+
```
Comment (thostetler) [inline src/pages/abs/[id]/exportcitation/[format].tsx:35]: should use the `useSettings` hook, but turn off suspense in the options passed

src/pages/search/exportcitation/[format].tsx
```diff
@@ -48,7 +47,12 @@ const ExportCitationPage: NextPage<IExportCitationPageProps> = (props) => {
     enabled: isAuthenticated,
   });
 
-  const settings = useMemo(() => settingsData ?? DEFAULT_USER_DATA, [settingsData]);
+  // fill any missing user data with default
+  const settings = useMemo(
+    () =>
+      settingsData ? { ...DEFAULT_USER_DATA, ...Object.entries(settingsData).filter((s) => !!s) } : DEFAULT_USER_DATA,
+    [settingsData],
+  );
```
Comment (thostetler) [inline src/pages/search/exportcitation/[format].tsx:55]: Same here, should probably use the `useSettings` hook

thostetler (commented):
Yeah I think this is the idea.  But we should consolidate any changes into the `useSettings` hook and just use that everywhere.
Make sure you turn off/on suspense as needed (right now it's on by default for the settings pages)

PR #407, opened 2024-01-03
Scix 351 libraries testing

shinyichen:
E2E tests for libraries landing page, library settings page, and library entity page.


e2e/libraries.entity.spec.ts
thostetler, line 1: I think you can probably just put all these inside a directory under `e2e`
`e2e/libraries/entity.spec`
Or we could do camel case, doesn't matter.

e2e/libraries.entity.spec.ts
```diff
@@ -0,0 +1,70 @@
+// back button should go to whereever it came from
+import { expect, test } from '@playwright/test';
+
```
thostetler, line 3: You can try to add a config line here to make this file run in parallel which might speed things up, locally anyway

```suggestion
test.describe.configure({
  mode: 'parallel',
});
```

thostetler (approved):
Looks good, could try to see if these will run well in parallel
And just to stay a little more consistent with the file naming scheme, maybe switch these to a `libraries` subdirectory

PR #429, opened 2024-02-07
Scix 296 libraries annotations

shinyichen:
A few new libraries functionalities:

### Filter by library type

https://github.com/adsabs/nectar/assets/636361/dbf47987-cc8c-40f9-b493-600ec7b6fc40

### Sort by name

https://github.com/adsabs/nectar/assets/636361/bb5cbab5-8d30-4a76-a0d4-bb1fedcdf7d5

### Annotation
Notes are globally visible to collaborators. Users with write permission can add and modify.

https://github.com/adsabs/nectar/assets/636361/c8d5f8fd-2d23-4b87-92dc-8187e2849520


thostetler (commented):
So I think overall these changes look good, and the new features work well.  I have a couple of things I wanted to mention.

- For all of the library lists there is some prop drilling.  I'd say if it becomes too onerous to update we should refactor into a top level data store and use context.  That may be better to cut down on rerenders, but I haven't noticed it being slow on my end so may be okay for now.

- For the annotations, I like the feature - but I'm not a fan of having a textarea be still shown in a *finalized* state.  Me personally, when I look at it I'm not sure if it's actually in EDIT mode or not.  I think it might be better to follow the same strategy github uses for example -- have a clean text output with a small "edit" button, then show a textarea so you can apply your edits.

- I'm not sure why the library list isn't caching when I switch between the different filters.  Might be something to look into, but probably out of scope for this PR.

All in all this is well done.  I'll just leave these as general feedback.

shinyichen:
@thostetler 

Thank you for the comments. I will work on the two suggestions.  And the library list is refetching because I intentionally disabled the cache, I wanted to have updated data after operations or changes made to library metadata. I think instead I can just `remove` cache whenever I need. I will make the changes

thostetler:
Makes sense, sounds good! 
Yeah probably after an operations change you can invalidate the caches.  
But it's hard since it may also get modified by a collaborator, so have to think about how to handle it best.

shinyichen:
@thostetler 
I've updated the UI for the annotation, and created a store. Please see if they look ok. Thanks.

src/pages/user/libraries/[[...id]].tsx
```diff
@@ -32,12 +40,20 @@ const LibrariesHome: NextPage<ILibrariesHomeProps> = ({ id, subpage, from }) =>
     { enabled: !!id, staleTime: 0 },
   );
 
+  useEffect(() => {
+    if (library) {
+      setLibrary(library);
```
Comment (thostetler) [inline src/pages/user/libraries/[[...id]].tsx:45]: I think actually we shouldn't do this.  Instead of using a store or prop drilling just refetch it in the child components.  Anything that we can get from the server, we should use react-query. 

Could definitely use this store though to manage the state for `onNoteUpdated` or `selectedBibcodes`, for example, which are currently getting passed down through.

thostetler (commented):
Changes look good, but I think the store could be improved, instead of setting anything from the store --- just use it to help handle the state for library table.  But like I mentioned before it may be fine, I haven't noticed any slow down

shinyichen:
I've updated the code so that I am mostly passing down the library id, and have components fetch the data. So I removed the store.

thostetler (approved):
Nice, thanks for the changes -- this looks good to go

PR #431, opened 2024-02-08
SCIX-366 Fix login weird redirect

thostetler:
- Fix redirect issue with login

### DIFF
--- a/src/components/Notification/Notification.tsx
+++ b/src/components/Notification/Notification.tsx
@@ -1,5 +1,5 @@
 import { useToast } from '@chakra-ui/react';
-import React, { useEffect } from 'react';
+import React, { useCallback, useEffect } from 'react';
 import { useStore } from '@store';
 import { useRouter } from 'next/router';
 import { NotificationId } from '@store/slices';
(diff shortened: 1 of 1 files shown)


shinyichen:
Here are two scenarios the redirect behavior after logging in:

Case 1: I am redirected to the landing page with the toast 'success' displayed. The url at this point is `/?notify=account-login-success`. After a few seconds I am redirected to `/`. 

Case 2: I am redirected to the landing page with the toast 'success' displayed. The url at this point is `/?notify=account-login-success`. Before the toast is closed, I go to the a different page (libraries) page, the new page is loading and then I am redirected back to the landing page again `/`.     ---> I think the redirect shouldn't happen in this case

thostetler:
@shinyichen yep you're right I see that behavior.  I think probably best to not bother trying to update the URL.
I think what's happening is that when the toast handler is loaded it is passed an `onClose` handler, this has it's current context as the `router.asPath: '/'` -- but when the page changes, the toast doesn't reload or get destroyed; and so when it finally closes it calls the handler which is now stale causing the redirect.

shinyichen:
Just tested the update, I am not able to close the toast message after logging in or logging out in. I have to navigate to another page (toast message is still there) then I am able to close it.

thostetler:
@shinyichen thanks for checking it out, I think I fixed this issue finally.  Just simplified the control further and made it react to route changes.  Please try again when you have a moment.  Thanks!

shinyichen:
Yes it's working for me now.

shinyichen (approved):

PR #432, opened 2024-02-09
Feedback form bibcode

shinyichen:
Only require bibcode in feedback record form if modifying existing record.


thostetler (approved):
These changes LGTM

PR #436, opened 2024-02-27
Fix/Mitigate Cascading Requests Issue

thostetler:
- **Fix unnecessary requests issue**
Disables react-query retrying failed requests for now


thostetler:
@shinyichen you'll need to add `NEXT_PUBLIC_API_HOST_CLIENT` to your `.env.local` file -- pointing to devapi

This will be used during development since the publicRuntimeConfig doesn't seem to get evaluated.

PR #449, opened 2024-03-19
Scix 373 Email Notification (formerly myADS) 

shinyichen:
https://github.com/adsabs/nectar/assets/636361/12f09739-448a-4dab-ab42-e36087049a9a


src/components/EmailNotifications/Forms/KewordsForm.tsx
```diff
@@ -0,0 +1,116 @@
...
+      editNofication(
+        {
+          id: notification.id,
+          data: keywords,
+          name,
+        },
+        {
+          onSettled(data, error) {
+            if (error) {
+              toast({ status: 'error', title: 'Error', description: parseAPIError(error) });
```
thostetler, line 48: This should probably be handled slightly differently.  Currently the description is not actually getting parsed because it's path is different than the other microservices: (`response/data/msg`) instead of `message` -- so we see: 
image
If I fix the parsing: 
image

The error message is worse: 
image

I think, on error here we should probably check if it's syntax issue and alert on the form instead.

thostetler (changes requested):
This functionally looks good to me.  I have a small concern about cleaning user input and I think we should improve the error handling a bit.

src/components/EmailNotifications/Forms/KewordsForm.tsx
shinyichen: @thostetler Please see my last commit. I added a syntax check.

shinyichen:
@thostetler Updated to check the validity of keyword. Let me know if it looks good to go.

thostetler (approved):
This is looking good to me

PR #456, opened 2024-03-23
SCIX-311 default collection not applied

thostetler:
Adds `All` to the default databases.
Gets default databases working on new searches

Improves logging and error handling in query-utils


src/pages/user/settings/application.tsx
```diff
@@ -100,20 +101,35 @@ const AppSettingsPage = () => {
...
-        v.value = true;
+  const handleApplyDatabases = useCallback(
+    (names: string[]) => {
+      const currentDatabases = settings[UserDataKeys.DEFAULT_DATABASE];
+
+      // if ALL is selected, reset to default (except for All)
+      if (names.includes(DatabaseEnum.All)) {
+        return setParams({
+          [UserDataKeys.DEFAULT_DATABASE]: [
+            ...DEFAULT_USER_DATA[UserDataKeys.DEFAULT_DATABASE],
```
shinyichen, line 112: The default data already has `{ name: DatabaseEnum.All, value: false }`, so here the result is 
```[
  {
    "name": "All",
    "value": false
  },
  {
    "name": "Physics",
    "value": false
  },
  {
    "name": "Astronomy",
    "value": false
  },
  {
    "name": "General",
    "value": false
  },
  {
    "name": "Earth Science",
    "value": false
  },
  {
    "name": "All",
    "value": true
  }
]```

shinyichen (changes requested):
A question: is the 'All' option necessary? Is it just convenience? Or does All include things beyond the specified categories? 

If I selected all categories but not 'All', is it equivalent to selecting 'All' ?

src/pages/user/settings/application.tsx
thostetler: Yep good catch, this should be fixed

thostetler:
> A question: is the 'All' option necessary? Is it just convenience? Or does All include things beyond the specified categories?
> 
> If I selected all categories but not 'All', is it equivalent to selecting 'All' ?

I agree, I think it's just a way of explicitly saying you want to NOT apply any filters.  
And yeah AFAIK all of these are the same:
- ALL selected (resulting in no collections applied)
- None selected (resulting in DEFAULT -- but in this case none  -- collections applied)
- All collections selected (resulting in an OR'd filter including all the collections)

I think the point is to allow users to explicitly set their intention here.  In Nectar, right now we don't have any global default filters, so it doesn't change anything to set `ALL` -- but if we did, then this would explicitly say we want NO filters applied.

shinyichen (approved):

PR #457, opened 2024-03-25
Library related fixes

shinyichen:
* Document item full journal name
* Hide relevance sort 
* Show an annotation icon for each document (recommended by Alberto, but other style suggestions welcome)
* Fix library documents not sorting


thostetler (approved):
LGTM

PR #467, opened 2024-04-04
Update abstract sources UI

shinyichen:
Update UI for abstract full text and other sources :
* In desktop, it is moved to the left side, and is always visible while navigating within the article
Screenshot 2024-04-05 at 4 13 11 PM

* In mobile, it is unchanged
Screenshot 2024-04-05 at 4 14 18 PM


shinyichen:
> I'm not sure how I feel about this menu pushing the navigation down, but it does work to make it more prominent on the page. If Kelly is cool with this then I'm good too.

I feel the same, but Alberto came up with it, and it might be better than the other designs.

PR #468, opened 2024-04-11
Scix 407 lib sort date added

shinyichen:
Sort library documents by date added, this is the default sort for library documents

Screenshot 2024-04-11 at 2 17 39 PM


thostetler (approved):
This looks okay to me, I think as an alternative we could pull out the list (and types) of the component and make it the parent's responsibility.  But this solution looks okay for now, and is simpler in this case, if a little less flexible.

shinyichen:
> This looks okay to me, I think as an alternative we could pull out the list (and types) of the component and make it the parent's responsibility. But this solution looks okay for now, and is simpler in this case, if a little less flexible.

Agreed. I didn't like the idea of mixing biblib sort with the rest of the solr sorts, but didn't want to mess up the current setup. I will commit this first, and start working on the refactor as well.

PR #469, opened 2024-04-17
Refactor sort drop down

shinyichen:
Refactor sort drop down. Let parents supply the sort list.

### DIFF
--- a/src/components/Sort/Sort.tsx
+++ b/src/components/Sort/Sort.tsx
@@ -1,29 +1,25 @@
-import { BiblibSortField, SolrSortDirection, SolrSortField } from '@api';
+import { SortDirection, SolrSortField, SortField, SortType } from '@api';
 import { Box, HStack, IconButton, Input } from '@chakra-ui/react';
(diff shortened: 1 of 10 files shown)


src/components/Sort/Sort.tsx
```diff
@@ -39,46 +35,34 @@ export interface ISortProps {
  *
  * Expects to be controlled (i.e. using sort and onChange to control value/updating)
  */
-export const Sort = (props: ISortProps): ReactElement => {
+export const Sort = <S extends SortType, F extends SortField>(props: ISortProps<S, F>): ReactElement => {
```
thostetler, line 38:
```suggestion
export const Sort = <S extends SortType = SolrSort, F extends SortField = SolrSortField>(props: ISortProps<S, F>): ReactElement => {
```

We could make solrSort the default so you don't have to pass the generic everywhere

thostetler (approved):
LGTM -- I think this makes for a cleaner component
:+1:

PR #474, opened 2024-04-27
Update path aliases and move logger

thostetler:
This changes the root path alias from `@` to `@/` which shows more explicitly project files vs external libs that might use that same syntax


src/middlewares/initSession.ts
```diff
@@ -141,10 +141,11 @@ export const initSession = async (req: NextRequest, res: NextResponse) => {
   if (isValidToken(token)) {
     session.token = token;
     session.isAuthenticated = isAuthenticated(token);
-    res.cookies.set(process.env.ADS_SESSION_COOKIE_NAME, headers.get('set-cookie') ?? '');
+    log.debug({ msg: 'headers', headers: headers.entries() });
+    const sessionCookieValue = headers.get('set-cookie').split('=')[1] ?? '';
```
shinyichen, line 145: possible `undefined` value on `headers.get('set-cookie')`?

shinyichen (approved):
Just that one comment to confirm no `undefined` error

PR #477, opened 2024-05-09
clear docs on search submission

thostetler:
On search submission (from the search bar) clear the selected docs


shinyichen (commented):
I think clear should also be called after applying filters, since some of the selected docs might not be in the filtered results.

thostetler:
> I think clear should also be called after applying filters, since some of the selected docs might not be in the filtered results.

hmm you might be right, maybe clear them for all changes except changing page

thostetler:
@shinyichen Updated this. Please check out when you get a chance, thanks

shinyichen (approved):

PR #478, opened 2024-05-20
Scix 182 facet accessibility

shinyichen:
Two parts:
* Part 1:  lots of small accessibility related fixes based on WAVE report
* Part 2: Add keyboard navigation to the search facet 
## Part 1
* Aria label
* Removed aria-labeledby in most <FormControl> since <FormLabel> sets htmlFor for children input. Except for the Switch component which requires a label.
* Fixed contrasts
* Improved keyboard accessibility
## Part 2
Add keyboard navigation for search facet.
* up and down arrows to navigate through the tree
* space to select
* right and left arrows to expand and collapse
* tab to jump to sections and buttons

I am using context to share current keyboard focused item, children count and expand state for all items.
I am using `facetId-rootIndex[-childIndex]` format as id for each item. 

This navigation doesn't apply when opened in modal, tab used in modal.


src/components/SearchFacet/SearchFacet.tsx
```diff
@@ -133,9 +143,23 @@ export const SearchFacet = (props: ISearchFacetProps): ReactElement => {
     padding: isDragging ? '4px' : undefined,
   };
 
+  // keyboard focus search facet and item
+  const [keyboardFocus, setKeyboardFocus] = useState<KeyboardFocusItem>({ facetId: null, index: [] });
```
thostetler, line 147: I'm thinking that this is probably defeating the purpose of the context you create below. Since this is causing the whole tree to get rerendered when we change focus.  

https://github.com/adsabs/nectar/assets/6970899/7ae5554e-68aa-4cad-973b-bb65417aed54

Probably for these list sizes it's not making a huge difference, however.  I think if wanna be efficient about it, this needs to be moved down to the item level and have the component watch the store for changes in focus state

src/components/SearchFacet/SearchFacet.tsx
```diff
@@ -209,9 +235,13 @@ export const SearchFacet = (props: ISearchFacetProps): ReactElement => {
           borderBottomRadius="md"
           mt="0"
         >
-          <FacetStoreProvider facetId={storeId} key={JSON.stringify(searchQuery)}>
-            <FacetList noLoadMore={noLoadMore} onFilter={handleOnFilter} onError={handleOnError} />
-          </FacetStoreProvider>
+          <FacetContext.Provider
```
thostetler, line 238: Probably unnecessary, I'd just add these to the current store.  Plus zustand provides us with ability to selective watch properties

src/components/SearchFacet/FacetList.tsx
```diff
@@ -128,14 +164,96 @@ export const NodeList = memo((props: INodeListProps) => {
     }
   };
 
+  const handleKeyboardFocusNext = (index: number[]) => {
```
thostetler, line 167: All this logic should probably be moved into the facetStore, and just have action calls on keyboard events.

src/components/SearchFacet/FacetList.tsx
```diff
@@ -128,14 +164,96 @@ export const NodeList = memo((props: INodeListProps) => {
...
       <List w="full" data-testid={`search-facet-${level}-list`} pl={level === 'child' ? 4 : 0}>
-        {treeData?.map((node) => (
-          <Item node={node} key={node.id} onError={onError} expandable={expandable} />
+        {treeData?.map((node, index) => (
+          <Item
+            node={node}
+            key={node.id}
+            onError={onError}
+            expandable={expandable}
+            facetId={facetId}
```
thostetler, line 242: Unless I'm mistaken you should be able to pull facetID from the store

thostetler:
So far looking pretty good, I noticed for the facets that the modal doesn't have the same keyboard navigation. 

I'm open to suggestions about best way to handle the facet state, but I think probably the best thing to do would be to move all the state into the existing store and add actions/handlers for it there.

thostetler (approved):
Changes look alright to me

PR #486, opened 2024-06-13
fix operations on selected papers

shinyichen:
* The useCallback needs dependencies
* Clear selected papers after applying operators


thostetler (approved):
LGTM

PR #493, opened 2024-07-03
fix orcid namevariations undefined

shinyichen:


thostetler (approved):
Changes look okay to me

PR #495, opened 2024-07-08
Fix abstract query page param

shinyichen:
Fix abstract query pagination param


thostetler (approved):
Changes look good to me

PR #496, opened 2024-07-09
Fix facet filter not applied

shinyichen:
Fix bug where only first search filter is applied. The query doesn't work correctly when the format of the array `fq` param is `comma`, but works if format is `repeat`.


thostetler (approved):
:+1:

PR #504, opened 2024-07-29
remember carousel position

shinyichen:
Use localStorage to keep carousel position

### DIFF
--- a/src/components/Pager/Pager.tsx
+++ b/src/components/Pager/Pager.tsx
@@ -47,9 +47,9 @@ const changePage =
       ret = noOfPages - 1;
     } else if (page === noOfPages - 1 && direction === 'next') {
       ret = 0;
+    } else {
+      ret = direction === 'next' ? page + 1 : page - 1;
     }
-    ret = direction === 'next' ? page + 1 : page - 1;
-
     onChange(ret);
 
     return ret;
--- a/src/components/Pager/Pager.tsx
+++ b/src/components/Pager/Pager.tsx
@@ -1,4 +1,4 @@
-import { forwardRef, HTMLAttributes, ReactNode, useState } from 'react';
+import { forwardRef, HTMLAttributes, ReactNode, useEffect, useState } from 'react';
 import {
   Box,
   ButtonProps,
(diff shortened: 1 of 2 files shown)


src/components/Pager/Pager.tsx
```diff
@@ -33,28 +34,43 @@ type Page = {
...
+  const [selectedPage, setSelectedPage] = useState(initialPage);
   const [interacted, setInteracted] = useState(false);
   const [shouldWiggle, setShouldWiggle] = useState(false);
 
-  const next = () => setSelectedPage(changePage(pages.length, 'next'));
-  const prev = () => setSelectedPage(changePage(pages.length, 'prev'));
-  const handlePageChange = (page: number) => setSelectedPage(page);
+  useEffect(() => {
+    setSelectedPage(initialPage);
+  }, [initialPage]);
```
thostetler, line 66:
```suggestion
```

Since the `initialPage` is an initializing prop, we expect it to be static -- so I think we could remove this useEffect.

src/pages/index.tsx
```diff
@@ -104,8 +104,20 @@ export default HomePage;
 export { injectSessionGSSP as getServerSideProps } from '@/ssr-utils';
 
 const Carousel = () => {
+  const [initialPage, setInitialPage] = useState<number>(0);
+
+  useEffect(() => {
+    setInitialPage(parseInt(localStorage.getItem('carousel') ?? '0'));
```
thostetler, line 110:
```suggestion
    setInitialPage(parseInt(localStorage.getItem('carousel') ?? '0', 10));
```
Just explicitly setting the radix

thostetler (approved):
I think this looks good to me, If we end up using this carousel thing anywhere else we should probably incorporate this code into the component.

src/components/Pager/Pager.tsx
shinyichen: @thostetler I think I added this because initially this was set to default 0, and when client gets the saved page number from localStorage, it is updated.

src/pages/index.tsx
shinyichen: fixed

PR #505, opened 2024-08-01
use normalized citation count 

shinyichen:
use normalized citation count in result list and library doc list when sorted by normalized citation

Screenshot 2024-08-01 at 11 26 19 AM


thostetler (approved):
Changes lgtm

PR #506, opened 2024-08-02
update author count search suggestion

shinyichen:
https://github.com/user-attachments/assets/ee51b515-1b94-4fd0-8f25-ffc82ba3a043


thostetler (approved):
👌🏼

PR #507, opened 2024-08-05
custom export pagination

shinyichen:
Moved the pagination control to the top, for two reasons:
1. Custom format gets pagination control
2. It makes more sense next to the page information

Screenshot 2024-08-05 at 11 13 27 AM


thostetler (approved):
LGTM

PR #509, opened 2024-08-06
Show overview year graph by individual years

shinyichen:
Don't group the years together in Overview -> Years graph

Screenshot 2024-08-06 at 3 48 07 PM


thostetler (approved):
:+1:

PR #510, opened 2024-08-08
Scix 443 sorting preference

shinyichen:
* Add sorting preference in search settings
* Apply preferred sorting when searching from landing page. Sort is sticky if searching from results page. 
* Sort is default to 'score'


thostetler (approved):
LGTM

PR #514, opened 2024-08-27
improve page loading indicator

shinyichen:
Instead of moving the spinner as suggested, I removed the spinner but made the progress bar thicker. I hope this is more clear that the page is loading, but without the indicator in the way.

https://github.com/user-attachments/assets/6e1787a5-e455-4a5f-8553-ee050c395d36


thostetler (approved):
This works for me

PR #517, opened 2024-08-30
Scix 457 improve export citation max author

shinyichen:
Replaced some sliders with number input. For sliders with wide range of numbers, it is hard to some users to adjust to a specific number. Those are replaced with number input. 
* Max authors in citation export and settings
* Max authors cutoff in citation export and settings
* Number of records in citation export

Screenshot 2024-08-30 at 2 31 03 PM


shinyichen:
@kelockhart @thostetler 
This PR is based on a user reporting that she wasn't able to use the slider to precisely adjust to a specific number.  I don't have any problems dragging to any number I want, but it might be difficult for some with shaky hands using a mouse to stop at a precise number. I'd like to know if you think changing them to number input makes sense. The draw backs are there is no indication of min and max, so I added that to the description. And typing is in is slower than dragging a mouse, may  not be as convenient, but that's just me.

thostetler (approved):
This seems alright to me

kelockhart:
Yep, looks good. Where else does this switch apply? For example, if I click "switch to advanced mode", it reveals the "Maximum Authors Cutoff" slider - does this also get switched to the field mode?

PR #522, opened 2024-09-05
fix popover shifting on arrow key

shinyichen:
Fix a bug where the popover for the search bar shifts when using arrow keys.


thostetler (approved):
Nice

PR #544, opened 2024-10-04
volume content menu bug

shinyichen:
Fix bug that caused 'volume content' menu be enabled and disabled incorrectly.


thostetler (approved):
Looks good, you'll need to rebase/merge.  May not need to do the lockfile

PR #557, opened 2024-11-12
allow URL and DOI in feedback form associated articles related bibcodes

shinyichen:
* When relationship is 'other', allow URL and DOI in the associated bibcodes.
* DOI and URL are not validated.

Screenshot 2024-11-12 at 11 08 44 AM


thostetler (approved):
LGTM

PR #560, opened 2024-11-21
Journalsdb

shinyichen:
This is the initial journalsDB I did in the June retreat. We would like to have it in production for the UG meeting.

(Recording is coming, at the time I am writing this, devapi is broken)


thostetler (approved):
This looks good to me so far

PR #567, opened 2024-12-05
update planetary facet fields

shinyichen:
This has not been tested

Update UI to use "planetary_feature" and "planetary_feature_facet" fields instead of "gpn" and "gpn_facet" fields


thostetler (approved):
This looks good to me

PR #572, opened 2024-12-11
Citation bulk fetch

shinyichen:
Fetch citations in bulk

### DIFF
--- a/src/components/ResultList/SimpleResultList.tsx
+++ b/src/components/ResultList/SimpleResultList.tsx
@@ -1,10 +1,14 @@
 import { Flex, VisuallyHidden } from '@chakra-ui/react';
 import { useIsClient } from '@/lib/useIsClient';
 import PT from 'prop-types';
-import { HTMLAttributes, ReactElement } from 'react';
+import { HTMLAttributes, ReactElement, useMemo } from 'react';
 import { Item } from './Item';
 import { useHighlights } from './useHighlights';
 import { IDocsEntity } from '@/api/search/types';
+import { useGetExportCitation } from '@/api/export/export';
+import { useSettings } from '@/lib/useSettings';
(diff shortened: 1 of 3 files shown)


src/components/ResultList/SimpleResultList.tsx
```diff
@@ -38,6 +42,33 @@ export const SimpleResultList = (props: ISimpleResultListProps): ReactElement =>
...
+      bibcode: bibcodes,
+      sort: ['bibcode asc'],
+    },
+    { enabled: !!settings?.defaultExportFormat },
+  );
+
+  // a map from bibcode to citation
+  const defaultCitations = useMemo(() => {
+    const citationSet = new Map<string, string>();
+    if (!!citationData) {
```
thostetler, line 64: We should do better error checking here.  And possibly wrap with try/catch just in case

src/components/ResultList/SimpleResultList.tsx
```diff
@@ -61,6 +92,7 @@ export const SimpleResultList = (props: ISimpleResultListProps): ReactElement =>
           highlights={highlights?.[index] ?? []}
           isFetchingHighlights={allowHighlight && isFetchingHighlights}
           useNormCite={useNormCite}
+          defaultCitation={defaultCitations?.get(doc.bibcode)}
```
thostetler, line 95: Where does the `.get` come from? Is that something useMemo is adding to the return?

thostetler (changes requested):
Looks good, I think the one method needs to have better error handling.

src/components/ResultList/SimpleResultList.tsx
shinyichen: @thostetler I added try catch, and reset the map to empty if error.

thostetler (approved):
Looks good to me! 👍

PR #587, opened 2025-01-21
download facet list

shinyichen:
Add download button to facet modal, available only to single level right now.
Screenshot 2025-01-21 at 2 58 19 PM

### DIFF
--- a/src/components/SearchFacet/SearchFacetModal/SearchFacetModal.tsx
+++ b/src/components/SearchFacet/SearchFacetModal/SearchFacetModal.tsx
@@ -1,4 +1,4 @@
-import { ChevronLeftIcon, ChevronRightIcon } from '@chakra-ui/icons';
+import { ChevronLeftIcon, ChevronRightIcon, DownloadIcon } from '@chakra-ui/icons';
 import {
   Box,
   Button,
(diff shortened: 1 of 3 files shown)


src/components/SearchFacet/SearchFacetModal/SearchFacetModal.tsx
```diff
@@ -70,6 +75,32 @@ const ModalFacet = (props: ISearchFacetModalProps) => {
   const setLetter = useFacetStore((state) => state.setLetter);
   const { searchTerm, search, letter, sort } = useGetSearchTerm();
 
+  const toast = useToast({ duration: 2000 });
+
+  const downloadable = !params.field.endsWith('_hier');
+
+  const { treeData, isFetched, error } = useGetFacetData({
```
thostetler: Doesn't this ALWAYS do the request for the records when the modal loads?
I think this would be better to roll this download button into a component in this file and try to only make the request if the user clicks the button?

On a side note, it looks like the nodes don't get properly added.  After doing the 2k records request, it still makes requests as we paginate.  We should probably refactor to be able to handle cached nodes instead of relying on the react-query cache

thostetler (changes requested):

src/components/SearchFacet/SearchFacetModal/SearchFacetModal.tsx
shinyichen: @thostetler The 2k request has a different param (query key) from the single page requests, so react query is not going to use the cached 2k request. Now that we are not always fetching the 2k records until user clicks download, I wonder if we still need to hack into the cache. It seems like a rare case that a user downloads 2k records then starts to paginate.

thostetler: yeah agreed, I think the one-time download is probably rare enough to keep it as it's own request (non-cached)

thostetler (approved):

PR #588, opened 2025-01-22
Get first-author working on classic form

thostetler:
Makes sure that the classic form handles `^Smith,A` type first-author searches properly.

`^Smith, A  -->  first_author:("Smith, A")`

### DIFF
--- a/src/components/ClassicForm/helpers.ts
+++ b/src/components/ClassicForm/helpers.ts
@@ -22,12 +22,14 @@ import {
   length,
   lt,
   map,
+  partition,
   pipe,
   propEq,
   reject,
   replace,
   split,
   splitAt,
+  startsWith,
   T,
   tail,
   test,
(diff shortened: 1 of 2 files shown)


shinyichen:
I did a search on `^huchra, john` and resulted in an error message `There was an issue parsing the query`
The generated query was: `collection:(astronomy) author:() first_author:("huchra, john")`
I think it was the `author:()` causing the problem.

src/components/ClassicForm/helpers.ts
```diff
@@ -154,9 +156,36 @@ export const getLogic = (logic: LogicChoice): LogicChoice => {
...
+    const firstAuthorResult = isEmpty(firstAuthors)
+      ? undefined
+      : pipe(
+          map(replace(/^\^/, '')),
+          parsePrefixesAndWrapItems,
+          joinItemsWithLogic(logic),
+          wrapWithField('first_author'),
+        )(firstAuthors);
+
+    const otherAuthorResult = pipe(
```
shinyichen, line 173: Might need to check `isEmpty` here too.

Search `^huchra, john` generated this query: collection:(astronomy) author:() first_author:("huchra, john")

Getting error with this query because of the empty author list

shinyichen (changes requested):
Just the one bug otherwise looks good.

src/components/ClassicForm/helpers.ts
thostetler: Good eye! I think I missed that one.  I'll fix and merge

PR #593, opened 2025-02-06
add share public library links

shinyichen:
* User library page: add 'open as public library' for library that is public
* Public library page: add 'share' button to copy library link

Screenshot 2025-02-06 at 2 40 51 PM
Screenshot 2025-02-06 at 2 41 06 PM


thostetler (approved):
This looks good to me

PR #602, opened 2025-03-12
Make pagination ui look consistent

shinyichen:
Update all pagination UI to look consistent with the pagination from search results.

Screenshot 2025-03-12 at 2 28 08 PM


thostetler (approved):
LGTM

PR #604, opened 2025-03-13
Fix facet bug when an operator word is in the filter string

shinyichen:
Found this bug when applying UAT keyword filter 'galactic and extragalactic astronomy. The `and` was converted to capitalized and that seemed to turn into the 'and' operator and messed up the results. 

The fix: apply operator capitalization outside of quotes.

https://github.com/user-attachments/assets/fbe031e2-6ac4-4928-9ab8-becd0b8eaafb


thostetler (approved):
:+1:  nice!

PR #611, opened 2025-03-27
Move facet buttons to the side

shinyichen:
https://github.com/user-attachments/assets/b3219847-e249-47c7-b099-171701a2870a


thostetler (approved):
These changes look good to me

PR #620, opened 2025-04-22
Scix 380 citation helper

shinyichen:
* Take the first 100 records (if results/selected > 100) and use as citation helper input.
* Citation helper returns 100 (fixed) suggestions
* Add to library (if logged in)
* View as search results is sorted by date (original order is lost)

https://github.com/user-attachments/assets/4c2cfe57-d9c1-4a61-8bfd-f2ab44683dc0


thostetler (approved):
Looks okay to me.  If we find the load wait is a little too long we may want to change from SSR to CSR, but for now this seems to work alright

PR #621, opened 2025-05-06
Scix 569 uat autocomplete

shinyichen:


src/components/SearchBar/SearchInput.tsx
```diff
@@ -58,11 +61,33 @@ export const SearchInput = forwardRef<ISearchInputProps, 'input'>((props, ref) =
...
+      // convert to typeahead options and dispatch
+      const options = uatSearchTermData.uatTerms.map((r, i) => {
+        return {
+          value: `"${r.name}"`,
+          label: r.name,
+          desc: [r.name, ...(r.altNames || [])].join(', '),
+          id: i,
+          match: [] as string[],
+        } as TypeaheadOption;
+      });
```
thostetler, line 81: I wonder if this could be rolled into the API response so it gets cached, since you have to do it at some point, this would allow you to skip it for subsequent requests.

src/api/uat/uat.ts
```diff
@@ -0,0 +1,32 @@
...
+    ...options,
+  });
+};
+
+export const fetchUATTerms: QueryFunction<IUATTermsSearchReponse> = async ({ meta }) => {
+  const { params } = meta as { params: IATTermsSearchParams };
+
+  try {
+    const { data } = await axios.get(`/api/uat/${params.term}`);
+    return data;
```
thostetler, line 28: I'm wondering if you should convert to your TypeAhead items here since you'll need it in that format.

src/api/uat/uat.ts
```diff
@@ -0,0 +1,32 @@
...
+  } catch {
+    return { error: 'error fetching UAT terms' };
```
thostetler, line 30: Since it doesn't throw, does this properly act like an error to react-query?

src/api/uat/uat.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export const uatSearchKeys = {
+  term: (term: string) => [UATSearchKeys.TERM, term] as const,
+};
+
+export const useUATTermsSearch: ADSQuery<IATTermsSearchParams, IUATTermsSearchReponse> = (params, options) => {
+  return useQuery({
+    queryKey: uatSearchKeys.term(params.term),
+    queryFn: fetchUATTerms,
```
thostetler, line 17: I think we should set a cache time of around 30s so it doesn't fill up the cache for too long.  Probably the main case for caching is for a user who is editing a query (i.e. "s" -> "su" -> "sum" -> "su" -> "sun")

thostetler (approved):
Nice work on this! I think it works pretty well and you got the caching/debouncing working :)

I am wondering about the bloat in the searchInput file though, we should look into refactoring this some.  It's a few hundred lines long now.  We can try to do a refactoring pass at some point soon.

src/components/SearchBar/SearchInput.tsx
shinyichen: Initially I was doing just what you described. After thinking a bit more, I felt the API should not return something so specific, maybe later on it can be used in other components.

src/api/uat/uat.ts
shinyichen: I need to remove it, I followed your code for Bibstem, Bibstem catches error here for something else.

src/api/uat/uat.ts
shinyichen: 👌

src/api/uat/uat.ts
shinyichen: 👍

PR #626, opened 2025-06-11
use default sort order

shinyichen:
When user changes sort order, the default direction is used.
* Author count and first author, use asc
* Everything else, use desc

https://github.com/user-attachments/assets/155e33e5-4570-414f-9e09-8d13067b80e5


thostetler (commented):
:+1:

PR #629, opened 2025-06-20
Add banner for system wide message

shinyichen:
* If user is authenticated, use user data for last dismissed message
* Otherwise, use local storage

https://github.com/user-attachments/assets/533980bf-ee35-482c-a315-37aad0a025e7


thostetler (approved):
Changes LGTM

PR #632, opened 2025-07-03
Scix 590/search input defaults to when navigation to results page

thostetler:
Fixes the *:* issue when navigating between pages
Also implements SCIX-584


shinyichen:
@thostetler I test manually the *.* issue, I noticed that if I refresh the search results page (or open the URL directly), the search input is empty.

shinyichen (approved):

PR #635, opened 2025-07-14
Show download btn to all facet modals

shinyichen:
Show download button for all facet modals

### DIFF
--- a/src/components/SearchFacet/SearchFacetModal/SearchFacetModal.tsx
+++ b/src/components/SearchFacet/SearchFacetModal/SearchFacetModal.tsx
@@ -191,10 +191,10 @@ const FacetDownloadButton = () => {
     enabled,
   });
 
-  const pluckVal = (list: FacetItem[]) => pluck('val', list);
-  const cleanup = replace(/^0\//, ''); // remove prefix
-  const formatData = useCallback(() => pipe(pluckVal, map<string, string>(cleanup), join('\n'))(treeData), [treeData]);
-
+  const formatData = useCallback(
+    () => pipe<[FacetItem[]], string[], string[], string>(pluck('val'), map(replace(/^0\//, '')), join('\n'))(treeData),
+    [treeData],
+  );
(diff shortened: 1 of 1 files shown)


src/components/SearchFacet/SearchFacetModal/SearchFacetModal.tsx
```diff
@@ -211,23 +212,21 @@ const FacetDownloadButton = () => {
...
-          leftIcon={<DownloadIcon aria-hidden />}
-          isLoading={isFetching}
-        >
-          <Text>Download full list</Text>
-        </Button>
-      </Flex>
-    );
-  }
-  return null;
+  // if (isDownloadable) {
```
thostetler, line 215:
```suggestion
```

src/components/SearchFacet/SearchFacetModal/SearchFacetModal.tsx
```diff
@@ -192,7 +191,9 @@ const FacetDownloadButton = () => {
     enabled,
   });
 
-  const formatData = useCallback(() => pipe(pluck('val'), join('\n'))(treeData), [treeData]);
+  const pluckVal = (list: FacetItem[]) => pluck('val', list);
+  const cleanup = replace(/^0\//, ''); // remove prefix
+  const formatData = useCallback(() => pipe(pluckVal, map<string, string>(cleanup), join('\n'))(treeData), [treeData]);
```
thostetler, line 196:
```suggestion
  const formatData = useCallback(
    () => pipe<[FacetItem[]], string[], string[], string>(pluck('val'), map(replace(/^0\//, '')), join('\n'))(treeData),
    [treeData],
  );
```

Cleans this up some.  also `pluck` is automatically curried so you don't have to wrap it

thostetler (approved):
This looks good.  I noticed:
image
That when I click the download button it renders the component multiple times.

This probably isn't an issue -- but something to look into if we see some slow down in this component.  Unless you spot a specific change that caused this re-rendering

PR #638, opened 2025-07-28
Scix 603 fix search bar not focusing

thostetler:


src/components/SearchBar/models.ts
```diff
@@ -112,26 +112,33 @@ export const typeaheadOptions: TypeaheadOption[] = [
     match: ['reviews('],
     id: 26,
   },
+  {
+    value: 'uat:"',
```
shinyichen, line 116: `uat:"` missing the closing quote.

shinyichen (approved):
I just noticed the UAT suggestion is missing closing quote, otherwise looks good.

PR #642, opened 2025-08-05
Scix 588 export manifest

shinyichen:
Use the new endpoint to fetch export formats.


thostetler:
TypeError: Cannot read properties of undefined (reading 'value')
    at ExportMenu (webpack-internal:///(pages-dir-browser)/./src/components/ResultList/ListActions.tsx:834:81)
    at renderWithHooks (webpack-internal:///(pages-dir-browser)/./node_modules/.pnpm/react-dom@18.3.1_react@18.3.1/node_modules/react-dom/cjs/react-dom.development.js:15486:18)
...

image

Getting an error when loading the search results page originating from the ListActions component.
Looks like the hook needs to do some validation on the incoming format and default to some known one if it's not.

The PR looks good otherwise, I can take a look then after you've had a chance to replicate/fix

thostetler (approved):
LGTM

PR #649, opened 2025-08-20
fix(facets): Issue with alpha sorting in facets modal

thostetler:
In a previous change the field was hard-coded to `count`.  Not sure the reason, but this reverts that and properly passes the field to the facet fetcher.


shinyichen:
I'm actually getting error when sorting by alpha now. And the error is from 
https://github.com/adsabs/nectar/blob/master/src/components/SearchFacet/SearchFacetModal/AlphaSorter.tsx#L28

```Property 'getRadioProps' does not exist on type '{ state: RadioState; getCheckboxProps: PropGetter<Record<string, unknown>, DOMAttributes<DOMElement>>; getInputProps: PropGetter<...>; getLabelProps: PropGetter<...>; getRootProps: PropGetter<...>; htmlProps: {}; }'.```

PR #651, opened 2025-08-20
fix library pagination bug

shinyichen:
Fix bug where library items pagination's page selector is off by 1 after selecting a page. The pagination `page` is expecting the page number, but library entity component uses `page` as indexed from 0. To be consistent, I updated the library entity component to use `pageNum` instead of `page`, making it less confusing.


thostetler (approved):
Minor suggestion, but looks good to me

PR #659, opened 2025-08-27
Allow facet download at different levels

shinyichen:
* When user selects download full facet list, the current list/level shown is downloaded (instead of the root level).


thostetler (approved):
LGTM

PR #667, opened 2025-09-10
Scix 609/make paper form bibcode query smarter

thostetler:
Instead of always creating a bigQuery ID for any and all submission to the bibcode search on the paper form, this will create a simple query for smaller numbers of bibcodes.

This guarantees some speed up, since we don't need to send the extra vault request to get an id for use by the subsequent query.

### DIFF
--- a/src/pages/paper-form.tsx
+++ b/src/pages/paper-form.tsx
@@ -40,6 +40,8 @@ import { stringifySearchParams } from '@/utils/common/search';
 import { fetchVaultSearch, vaultKeys } from '@/api/vault/vault';
 import { SimpleLink } from '@/components/SimpleLink';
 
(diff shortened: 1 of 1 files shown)


shinyichen (approved):
Not sure how I would test it, but looks good from the code.

thostetler:
@shinyichen Thanks for checking it out!
Sorry I should have mentioned the steps to test

Essentially this makes a normal query if the amount of bibcodes in the paper form are under some threshold (e.g. 50). 

So you could go to a query export and use a custom export like `%R` to get bibcodes and then take some under and over the threshold to see how the paper form handles it differently.

-- 
Also, I'll fix the failing tests/lint issues

shinyichen:
@thostetler Tested and works!

PR #679, opened 2025-10-01
update bibstem to pub for search field

shinyichen:
* Fix 'pub' in the quickfield dropdown
* Fix 'pub' in auto-complete
* Update bibstem to pub in the search example

Screenshot 2025-10-01 at 8 57 50 AM
Screenshot 2025-10-01 at 8 58 09 AM


thostetler (approved):
👍

PR #832, opened 2026-03-31
Prepend DOI and ARXIV URL origin when adding new url in the form

shinyichen:
Avoid confusion of DOI and arxiv format in the URL by prepending the origin.

Screenshot 2026-03-31 at 2 25 50 PM
Screenshot 2026-03-31 at 2 27 15 PM

### DIFF
--- a/src/components/FeedbackForms/MissingRecord/UrlsField.tsx
+++ b/src/components/FeedbackForms/MissingRecord/UrlsField.tsx
@@ -24,6 +24,10 @@ const typeOptions: SelectOption<ResourceUrlType>[] = resourceUrlTypes.map((t) =>
   value: t as string,
 }));
 
+const DOI_ORIGIN = 'https://doi.org';
+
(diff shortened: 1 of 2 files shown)


src/components/FeedbackForms/MissingRecord/UrlsField.tsx
```diff
@@ -24,6 +24,10 @@ const typeOptions: SelectOption<ResourceUrlType>[] = resourceUrlTypes.map((t) =>
   value: t as string,
 }));
 
+const DOI_ORIGIN = 'https://doi.org';
```
thostetler: Looks good, however I wonder if there is a case for putting these in our central config `src/config.ts` ?

thostetler (approved):
LGTM -- just a quick question

src/components/FeedbackForms/MissingRecord/UrlsField.tsx
shinyichen: @thostetler updated, please check.

PR #835, opened 2026-04-03
Default sort search by orcid id by date

shinyichen:
Update sort to date when clicking on the orcid id search.


thostetler (approved):
Looks good

PR #837, opened 2026-04-07
Fix incorrect record form single url no loading

shinyichen:
Update how we get a record's esource (urls) in incorrect record form, so that all urls can be loaded.

Previous we were using link_gateway page and parse urls in the html page. This was a problem when a record has only one url, which was instead redirected to that url instead. We were not able to parse the single url.

The screenshots show that records with single url and multiple urls are loaded.

2013admf.book.....D
Screenshot 2026-04-07 at 12 47 29 PM

2022AAS...24034906L
Screenshot 2026-04-07 at 12 48 00 PM


thostetler (approved):
I think these changes look reasonable 👍

PR #840, opened 2026-04-08
fix(ui): sort height alignment, paper form button sizes, facet gap

thostetler:
- align sort dropdown height (40px) with adjacent asc/desc toggle button
- add missing `size="sm"` to Reset buttons on paper form (all three sub-forms)
- remove empty aside flex child when facets are hidden, eliminating left-side gap in results layout

image

image

image

### DIFF
--- a/src/pages/paper-form.tsx
+++ b/src/pages/paper-form.tsx
@@ -236,7 +236,7 @@ const JournalQueryForm = ({ onSubmit, error }: SubFormProps) => {
           <Button size="sm" isDisabled={isSubmitting} type="submit" isLoading={isSubmitting}>
             Search
           </Button>
-          <Button type="reset" variant="outline" onClick={handleReset} isDisabled={isSubmitting}>
+          <Button size="sm" type="reset" variant="outline" onClick={handleReset} isDisabled={isSubmitting}>
             Reset
           </Button>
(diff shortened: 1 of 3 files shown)


src/pages/paper-form.tsx
```diff
@@ -236,7 +236,7 @@ const JournalQueryForm = ({ onSubmit, error }: SubFormProps) => {
           <Button size="sm" isDisabled={isSubmitting} type="submit" isLoading={isSubmitting}>
             Search
           </Button>
-          <Button type="reset" variant="outline" onClick={handleReset} isDisabled={isSubmitting}>
+          <Button size="sm" type="reset" variant="outline" onClick={handleReset} isDisabled={isSubmitting}>
```
shinyichen: All the buttons on Nectar are set to default size (md, see theme file) so that it follows WCAG AAA guideline. I guess it's not an issue if they are set to SM here, if you feel it's appropriate. But if you want to stay consistent, you don't need to set the button size

shinyichen (approved):
Looks good, please see my comment.

src/pages/paper-form.tsx
thostetler: Yep agreed, would rather stick within our norms

PR #845, opened 2026-04-13
fix(libraries): prevent adding documents to read-only libraries

thostetler:
- Read-only libraries are visually disabled (greyed out, `aria-disabled`) in the **Add to Library** modal and cannot be selected — closes SCIX-864
- Switched `mutate` → `mutateAsync` so `Promise.all` actually awaits the API call; previously `mutate` returned `void`, causing `Promise.all` to resolve immediately and always show the success banner regardless of the 403 response
- MSW mock handler now returns 403 when attempting to add/remove documents on a library with `permission === 'read'`
- `permission` column is now visible in the modal table so users can see why some rows are disabled
- Added `disabledIds` prop to `LibraryListTable` for general-purpose row disabling
- Add tests

image

### DIFF
--- a/src/components/Libraries/AddToLibraryModal.tsx
+++ b/src/components/Libraries/AddToLibraryModal.tsx
@@ -234,7 +234,7 @@ const AddToExistingLibraryPane = ({
             showIndex={false}
             showSettings={false}
             showDescription={false}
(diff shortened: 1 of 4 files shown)


shinyichen (approved):
👍

PR #847, opened 2026-04-14
Scix 837 orcid UI update

shinyichen:
* Update the ORCiD profile panel
** Update the orcid id to link to user's orcid public profile page
** Add a button that launches search using user's orcid
* Fix footer text overflow in mobile view

Screenshot 2026-04-14 at 2 59 04 PM

Screenshot 2026-04-14 at 2 59 30 PM

Footer Before
Screenshot 2026-04-14 at 3 00 38 PM

Footer After
Screenshot 2026-04-14 at 3 00 53 PM


thostetler (approved):
Looks pretty good! I like the usage of the scix logo on the "search by" button -- good idea!

PR #848, opened 2026-04-17
use matching environment for token setting

shinyichen:
In token setting, use the matching env for the API usage code.

### DIFF
--- a/src/pages/user/settings/token.tsx
+++ b/src/pages/user/settings/token.tsx
@@ -153,15 +153,15 @@ const TokenArea = (props: { onGenerate: () => void; isLoading: boolean }) => {
           </Heading>
           <Stack direction={{ base: 'column', xs: 'row' }}>
             <Code>
-              {
-                "curl -H 'Authorization: Bearer *************************' https://api.adsabs.harvard.edu/v1/search/query?q=star&fl=bibcode"
-              }
+              {`curl -H 'Authorization: Bearer *************************' ${
+                process.env.NEXT_PUBLIC_API_HOST_CLIENT ?? 'http://scixplorer.org/v1'
+              }/search/query?q=star&fl=bibcode`}
             </Code>
             <SimpleCopyButton
               variant="solid"
-              text={`curl -H 'Authorization: Bearer ${
-                data?.access_token ?? ''
-              }' https://api.adsabs.harvard.edu/v1/search/query?q=star&fl=bibcode`}
+              text={`curl -H 'Authorization: Bearer ${data?.access_token ?? ''}' ${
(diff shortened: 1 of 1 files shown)


src/pages/user/settings/token.tsx
```diff
@@ -153,15 +153,15 @@ const TokenArea = (props: { onGenerate: () => void; isLoading: boolean }) => {
           </Heading>
           <Stack direction={{ base: 'column', xs: 'row' }}>
             <Code>
-              {
-                "curl -H 'Authorization: Bearer *************************' https://api.adsabs.harvard.edu/v1/search/query?q=star&fl=bibcode"
-              }
+              {`curl -H 'Authorization: Bearer *************************' ${
+                process.env.NEXT_PUBLIC_API_HOST_CLIENT ?? 'http://scixplorer.org/v1'
```
shinyichen: Is it a good idea to default to http://scixplorer.org/v1 here? Same thing on line 162

shinyichen:
@thostetler I wasn't able to test this because I am getting 401 when fetching token. Taylor hasn't figured out why either. If you are able to, can you confirm whether it is working?

src/pages/user/settings/token.tsx
thostetler:
```suggestion
                process.env.NEXT_PUBLIC_API_HOST_CLIENT ?? 'https://scixplorer.org/v1'
```

src/pages/user/settings/token.tsx
```diff
@@ -153,15 +153,15 @@ const TokenArea = (props: { onGenerate: () => void; isLoading: boolean }) => {
...
+              }/search/query?q=star&fl=bibcode`}
             </Code>
             <SimpleCopyButton
               variant="solid"
-              text={`curl -H 'Authorization: Bearer ${
-                data?.access_token ?? ''
-              }' https://api.adsabs.harvard.edu/v1/search/query?q=star&fl=bibcode`}
+              text={`curl -H 'Authorization: Bearer ${data?.access_token ?? ''}' ${
+                process.env.NEXT_PUBLIC_API_HOST_CLIENT ?? 'http://scixplorer.org/v1'
```
thostetler:
```suggestion
                process.env.NEXT_PUBLIC_API_HOST_CLIENT ?? 'https://scixplorer.org/v1'
```

src/pages/user/settings/token.tsx
thostetler: I think that should be okay

thostetler (approved):
LGTM, couple small changes

I was able to confirm that QA and DEV environments show the correct URLs
Image

PR #850, opened 2026-04-17
Fix record doesn't load in feedback form when title is undefined

shinyichen:
Define a default record title in incorrect record form when title is undefined. Since title is a required field in the form, an empty space is used as default, otherwise the form does not load properly.


thostetler (approved):
Nice

PR #856, opened 2026-04-21
Exit tour when clicked outside the tour's modal

shinyichen:
https://github.com/user-attachments/assets/56b6b89b-bd0c-41a4-a63e-812ba23ada61


thostetler (approved):
I think this looks reasonable to me

PR #860, opened 2026-05-05
SCIX-871 fix: gate facets on searchStatus to prevent stale data

thostetler:
Facets were showing stale data from the previous search when a new search returned zero results or errored. The root cause: latestQuery (what facets use to build their requests) only updated when docs.length > 0, and facets had no signal for whether the main search was still in flight, empty, or failed.

- Add SearchStatus union type to the Zustand search slice ('idle' | 'loading' | 'success' | 'empty' | 'error')
- Drive status transitions from the search page using isLoading/isSuccess/isError — uses `isLoading` rather than isFetching to avoid toggling facets off during background refetches of the same query
- Switch status effect to useLayoutEffect so facets start loading in the same frame as search results
- Gate facet fetches (enabled) and returned data (treeData, totalResults, canLoadMore) on searchStatus === 'success' — explicit data gate required because enabled: false alone does not clear React Query cached values
- Show a spinner in open facets while the main search is loading instead of incorrectly showing No Results
- Fix store singleton in test-utils that was leaking Zustand state between tests
- Fix TS errors in SearchFacetModal (children render prop type conflict) and FacetStoreProvider

### DIFF
--- a/src/components/SearchFacet/FacetList.tsx
+++ b/src/components/SearchFacet/FacetList.tsx
@@ -28,7 +28,6 @@ import {
   PopoverCloseButton,
   PopoverContent,
   PopoverHeader,
-  Skeleton,
   Spinner,
   Stack,
   Text,
@@ -122,7 +121,7 @@ export const NodeList = memo(
     const updateModal = useFacetStore(selectors.updateModal);
     const depth = getLevelFromKey(prefix) + 1;
     const expandable = params.hasChildren && (level === 'root' || params.maxDepth > depth);
-    const { treeData, isFetching, isLoading, isError } = useGetFacetData({
+    const { treeData, isFetching, isLoading, isSearchLoading, isError } = useGetFacetData({
       ...params,
       prefix,
       level,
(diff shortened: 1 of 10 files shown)


src/components/SearchFacet/FacetList.tsx
```diff
@@ -158,10 +157,10 @@ export const NodeList = memo(
       );
     }
 
-    if (isFetching || isLoading) {
+    if (isFetching || isLoading || isSearchLoading) {
       return (
-        <Center data-testid="search-facet-loading">
-          <Spinner size="sm" />
+        <Center py="4" data-testid="search-facet-loading">
+          <Spinner size="sm" color="gray.400" />
```
shinyichen: The hard coded color here might be an accessibility issue (although it doesn't stay on screen that long). I believe gray.400 doesn't have enough contrast on white background, and also considering we also have dark mode, maybe it's best to specify the color using the useColorModeColor hook.

src/components/SearchFacet/FacetList.tsx
```diff
@@ -313,20 +313,11 @@ export const NodeListModal = (props: INodeListProps) => {
...
-        <Skeleton h="24px" />
-        <Skeleton h="24px" />
-        <Skeleton h="24px" />
-        <Skeleton h="24px" />
-        <Skeleton h="24px" />
-        <Skeleton h="24px" />
-        <Skeleton h="24px" />
-      </Stack>
+      <Center py="4" data-testid="search-facet-loading">
+        <Spinner size="sm" color="gray.400" />
```
shinyichen: Same as above.

shinyichen (approved):

PR #861, opened 2026-05-05
SCIX-871 fix: gate SortStats and YearHistogramSlider on searchStatus

thostetler:
NumFound's SortStats sub-component and YearHistogramSlider both read latestQuery
directly and fire API requests when the main search has errored, showing stale
citation stats and histogram data from the previous query.

- Gate SortStats rendering in NumFound on searchStatus === 'success' so the
  citation stats call never fires on stale latestQuery
- Add searchStatus === 'success' to the enabled condition in YearHistogramSlider
  so histogram data is not fetched against a stale query


thostetler:
Folded into #860.

PR #864, opened 2026-05-08
fix(facets): export label and count columns in facet CSV download

thostetler:
Facet modal download only exported the label column — the count was never included it seems?

- Extracts `formatFacetCSV` as a pure helper that emits a proper two-column CSV (`Label,Count`) 
- Updates `FacetDownloadButton` to use the helper; changes filename to `fulllist.csv` and MIME type to `text/csv`
- Adds unit tests

Closes SCIX-872


thostetler:
@shinyichen should this one be CSV? 
I just noticed that these output as txt

Let me know if these changes makes sense.  Thanks!

shinyichen:
> @shinyichen should this one be CSV? I just noticed that these output as txt
> 
> Let me know if these changes makes sense. Thanks!

@thostetler Do you mean the search facet download file? It was just a list, so I didn't use CSV. But makes sense to make it CSV now.

shinyichen (approved):

PR #866, opened 2026-05-14
SCIX-867 feat: add SearchModifierBar with ADS Compatibility mode

thostetler:
The app had no way for users to scope searches to ADS-style astronomy and physics content with date-sorted results. ADS users arriving via referrer also had no awareness that their experience was configured differently.

- Add SearchModifierBar dropdown (All relevant content / ADS Compatibility mode) on the landing page when in Astrophysics mode
- Persist mode in Zustand store and scix_prefs cookie, seeded by middleware on ADS referrer detection and discipline route redirects
- Apply fq_database filter and date sort when ADS Compatibility mode is active; strip them on mode switch
- Show toast notification when arriving from ADS referrer (fromADS param)
- Submit-time mode application on landing and abstract form pages
- Strip ads_compat param from Solr query key to prevent double-fetch on AppMode sync
- SearchModifierBar is shown only on the landing page (ASTROPHYSICS mode) — search results page intentionally omitted (tentative product decision, may be revisited)

### DIFF
--- a/src/__tests__/ssr-utils.test.ts
+++ b/src/__tests__/ssr-utils.test.ts
@@ -16,6 +16,7 @@ const getMockContext = (
   query: ParsedUrlQuery = {},
   resolvedUrl = '/search',
   referer?: string,
+  cookie?: string,
 ) =>
   ({
     req: {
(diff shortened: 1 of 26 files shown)


shinyichen (approved):
👍

PR #867, opened 2026-05-14
Item Resource dropdown: accessibility and layout fixes

shinyichen:
* The tooltip was causing the dropdown menus to be no longer accessible by keyboard. 
  * Move the tooltip from the menu to the icon 
* Fix inconsistent menu icons size and layout

Before:
Screenshot 2026-05-14 at 3 00 48 PM

After:
Screenshot 2026-05-14 at 3 01 04 PM


thostetler (approved):
LGTM

PR #871, opened 2026-05-27
SCIX-846 Fix List Components in Feedback Forms

thostetler:
Feedback forms that use a staging input pattern (type a value, click + to commit) silently
  discard the staged value when users blur away without clicking the button. Single-entry
  users who type one value and move directly to Preview find the form stuck with the button
  disabled and no explanation.

  - Add container-level onBlur to the new-bibcode group in AssociatedTable, the new-URL row in
   UrlsTable, and the new-reference row in MissingReferenceTable
  - Auto-commit the staged value when focus leaves the container, if the value is valid
  - Guard with e.currentTarget.contains(e.relatedTarget) to skip auto-commit when focus moves
  between siblings within the same row, preventing double-commits when the user clicks the
  explicit Add button
[Kooha-2026-05-27-14-16-52.webm](https://github.com/user-attachments/assets/7c435195-7506-45c1-9617-bb9f37406619)


shinyichen (approved):
👍

PR #876, opened 2026-05-28
Convert export page to be CSR

thostetler:
Export page has been extremely slow to load, and since most of the wait is done server-side waiting on the API to respond we can offload that to the client to take advantage of the RC caching and being able to show skeletons.

This also removes the infinite style calls, which probably aren't necessary here.


shinyichen (commented):
* I wasn't able to switch format using the dropdown. 
* After the skeletons disappear, the export text area first says "press submit to generate report", after a few seconds, the results are loaded. We probably should show the loading spinner just like when we do switch to a different format (check scixplorer.org since format switch is not working now).

https://github.com/user-attachments/assets/2c716633-de35-47fa-9c21-7de2d1e91f16

PR #890, opened 2026-06-18
SCIX-881 refactor(search): centralize back-to-results navigation

thostetler:
Back-to-results controls all called history.back(), which breaks when in-page navigation (export format changes, sort/filter/paging) pushes history entries between the results page and the current one. Abstract subpages had no search query in their URL, so no return target at all.

1. Add useSearchReturnTo: resolves a return URL by precedence — library referrer, captured session URL, page-reconstructable URL, none.
2. Capture the results URL per tab in sessionStorage (client-only, after the route settles, results views only) via a small session storage utility.
3. Route SearchBar, author affiliations, citation helper, export, and the viz pages through one BackToSearchResults control, hidden when nothing resolves.
4. CitationExporter format changes use router.replace instead of push.

Left CitationExporter.beforePopState alone; removing it is a separate change.


shinyichen (approved):
One thing I think can be updated (here or later) is that when clicking on an abstract in a library, the "back to search results" makes more sense to be "back to library" and links back to the library.

thostetler:
Thats a good idea, may be worth doing here -- let me see how hard that would be to implement

--- edit
I'm gonna push that for a follow-up

PR #892, opened 2026-06-22
add volume and page in search result items

shinyichen:
Screenshot 2026-06-22 at 10 40 31 AM

### DIFF
--- a/src/components/ResultList/Item/Item.tsx
+++ b/src/components/ResultList/Item/Item.tsx
@@ -7,7 +7,7 @@ import {
   Flex,
   Stack,
   Text,
-  Tooltip,
+  useBreakpointValue,
   useTimeout,
 } from '@chakra-ui/react';
 import { AuthorList } from '@/components/AllAuthorsModal';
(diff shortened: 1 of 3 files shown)


src/components/ResultList/Item/Item.tsx
```diff
@@ -163,23 +164,30 @@ export const Item = (props: IItemProps): ReactElement => {
...
+          <Stack
+            direction={{ base: 'column', md: 'row' }}
+            fontSize="xs"
+            mt={0.5}
+            gap={{ base: 0.5, md: 0 }}
+            flexWrap="wrap"
+            divider={divider}
+          >
+            <Text>{formattedPubDate}</Text>
+            <Text>{truncatedPub}</Text>
```
thostetler:
```suggestion
            <Tooltip
              label={pub}
              aria-label="publication tooltip"
              placement="top"
              isDisabled={!pub || pub.length <= APP_DEFAULTS.RESULT_ITEM_PUB_CUTOFF}
            >
              <Text>{truncatedPub}</Text>
            </Tooltip>
```

Wonder if we should allow expanding the full pub name for when it gets truncated?

thostetler (approved):
Looks great to me, I just had the one suggestion

src/components/ResultList/Item/Item.tsx
thostetler: this would need to be duplicated in the library too

shinyichen: 👍

PR #902, opened 2026-07-17
Scix 883 updated tour

shinyichen:
When the discipline is Astrophysics (either user selected or redirected from ADS), the tour is the ADS user version. 
Screenshot 2026-07-17 at 1 57 04 PM


thostetler (changes requested):
Looks pretty good.

One thing, I think we should prefer something like `data-tour-id` over using `id` just so we don't break something if we add some accessibility linking or whatever.  
Also it's more clear in the dom elements which ones are tour landmarks.

shinyichen:
@thostetler I've updated the `id` field to `data-tour`. Please see my update.

thostetler (approved):
👍

PR #905, opened 2026-07-24
SCIX-890 fix(vis): keep paper limit and CSV reachable on empty viz

thostetler:
The author and paper network pages dropped the entire graph pane when the
backend couldn't build groups, so a user who over-restricted the paper limit
had no way to raise it back and nothing to export. The "not enough data"
response also uses a different shape than success (no root/summaryGraph; raw
rows under fullGraph), which the CSV path wasn't reading.

1. Add a shared NotEnoughData panel that keeps the paper limit control on
   screen and offers an optional CSV download.
2. Build the error-state CSV from fullGraph.nodes (author name and weight;
   paper bibcode and metadata), omitting the button when there are no rows.
3. Model the fullGraph fallback in the vis response types.
image

### DIFF
--- a/src/api/vis/types.ts
+++ b/src/api/vis/types.ts
@@ -12,8 +12,27 @@ export interface IADSApiWordCloudParams {
 }
 
 export interface IADSApiAuthorNetworkResponse {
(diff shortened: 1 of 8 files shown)


shinyichen (approved):
👍

PR #908, opened 2026-07-30
SCIX-893 feat(search): add title quickfield and all-terms modal

thostetler:
The quickfields row lacked a title field, and its all-terms combobox used a cramped 200px input with a popper tooltip that was itself hidden below md.

- Add title as the trailing quickfield in both default and biology/physical-science modes
- Replace the combobox with an icon-button trigger that opens a command-palette modal: filter, grouped browse list, keyboard nav, and a detail pane with sanitized description, syntax, and examples
- Reveal quickfields by breakpoint so trailing fields drop first on small screens (e.g. author and first author stay, the rest appear as the row widens); the trigger label collapses to an icon there too
- Derive the quickfields from the shared term list so field metadata has a single source of truth
- Preserve the quick-fields and all-search-terms tour hooks

Ticket: SCIX-893 and SCIX-894.
demo: [Kooha-2026-07-30-16-29-43.webm](https://github.com/user-attachments/assets/400979aa-2243-48f1-8757-f6bb1fb8b6c0)


shinyichen (commented):
Looks good. I just want to point out that descriptions contain links, and with a mouse it is very hard to get to the link. If I move my mouse down towards the description area, the description would have moved to another term. I could move my mouse outside the modal, than down than towards the description, but we shouldn't expect users to do that. I wonder if the description can be moved to the right side to resolve that problem.

https://github.com/user-attachments/assets/dbe6bbe6-6051-4a4c-ac1e-b318f7846193

thostetler:
@shinyichen good point, let me think about how to handle links here

thostetler:
[Kooha-2026-08-05-08-36-07.webm](https://github.com/user-attachments/assets/643c54aa-8cd7-4fcd-bc28-11d0b84166f5)
Shifted it to the right like you mentioned, I think this two-column setup is a little better.  It should shift to be one-column on mobile

thostetler:
@shinyichen if you wouldn't mind, please check this out one more time.  Thanks!

shinyichen (approved):
Looks awesome!

PR #909, opened 2026-07-31
size up add to library modal for better layout

shinyichen:
Make add to library modal larger to accommodate long library names

 
Screenshot 2026-07-31 at 10 49 00 AM


thostetler (approved):
Nice

PR #911, opened 2026-08-04
SCIX-900 fix(orcid): auto-expire ORCiD mode after 5 minutes of inactivity

thostetler:
ORCiD mode stayed on indefinitely once toggled — no timeout, and it
persists across sessions via localStorage.

1. lastActivityAt on the orcid slice, stamped by toggle-on, claim/
   update/remove work, and /user/orcid visits
2. single root-mounted timer (avoids duplicate timers across the
   several useOrcid instances) turns mode off and toasts on expiry
3. stale mode cleared silently on rehydration, including legacy
   persisted state with no lastActivityAt

image


shinyichen (approved):
👍

PR #913, opened 2026-08-04
Improve library operation modal

shinyichen:
* Add operation description
* Updated 'difference' operation form to differentiate source and secondary libraries

Screenshot 2026-08-04 at 12 24 28 PM


thostetler (approved):
Looks pretty good to me!

